PR #2, opened 2017-07-06
Readme

mbrandonw:
just a basic readme...

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ let package = Package(
 
 ## Table of Contents
 
-* [`HTML`](#HTML)
-* [`CSS`](#CSS)
-* [`HTTPPipeline`](#HTTPPipeline)
-* [`ApplicativeRouter`](#ApplicativeRouter)
+* [`HTML`](#html)
+* [`CSS`](#css)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,2 +1,140 @@
...
+
+This library should be considered alpha, and not stable. Breaking changes will happen often.
+
+## Installation
+
+```swift
+import PackageDescription
+
+let package = Package(
+    dependencies: [
```
stephencelis: Should this be 2-space-indent for consistency?

README.md
```diff
@@ -1,2 +1,140 @@
...
+```text
+Status 200
+Content-Type: text/plain
+
+Hello, world
+```
+
+## `ApplicativeRouter`
+
+A router built on the principles of “applicative parsing” that is robust, composable and type-safe. It’s job is to take the incoming, unstructured `URLRequest` from the browser and turn it into a structured value so that your app can do what it needs to do to produce a response.
```
stephencelis: "Its job" 🎓

mbrandonw: lol

PR #4, opened 2017-07-08
Added a pretty printer package

mbrandonw:
This adds a whole new package `import HtmlPrettyPrint` that uses @bkase's [DoctorPretty](https://github.com/bkase/DoctorPretty.git) lib to do a prettier render of HTML. It was pretty easy to set up, and will be super useful for tests!


bkase:
Looks great!

PR #5, opened 2017-07-09
Prettier printing for class and style attributes

mbrandonw:
This adds special logic around rendering `class` and `style` attributes in html nodes. it's common for those attributes to have very long values, and so it would be nice to give custom logic on how to insert line breaks.


Sources/HTML/HTML.swift
```diff
@@ -30,6 +30,7 @@ public struct Attribute {
 }
 
 public protocol Value {
+  func renderedValue() -> EncodedString?
```
mbrandonw: so I had to add this so that I could better customize how certain attributes were rendered. it's a bit hacky, but I think we could clean it up easily later.

Tests/HtmlPrettyPrintTests/PrettyTests.swift
```diff
@@ -34,8 +38,15 @@ class PrettyTests: XCTestCase {
...
-         style="color: red;">
+         class="foo
+                bar
+                baz
+                class1
+                class2
+                class3"
+         style="color: red;
+                background: blue;
+                padding: rem(2);">
```
mbrandonw: p cool! had these styles/classes fit on one line there would be no newlines. but once they break, they all align together.

stephencelis: whooa

PR #9, opened 2017-07-14
Pretty print text

mbrandonw:
Realized I should have been pretty printing `text` nodes so that they wrap around properly.


Tests/HtmlPrettyPrintTests/HtmlPrettyPrintTests.swift
```diff
@@ -22,7 +22,11 @@ class PrettyTests: XCTestCase {
                   ],
                   ["hello world"]
                 ),
-                p(["goodbye world"]),
+                p([
+                  """
+Tacit programming, also called point-free style, is a programming paradigm in which function definitions do not identify the arguments (or "points") on which they operate.
```
stephencelis: lol

PR #10, opened 2017-07-16
Some random clean up

mbrandonw:
Just some things I did while writing up this pretty printing article...


Sources/HtmlPrettyPrint/Pretty.swift
```diff
@@ -33,34 +33,37 @@ private func prettyPrint(node: Node) -> Doc {
...
   return .text("<")
     <> .text(element.name)
     <> prettyPrint(attributes: element.attribs)
-    <> (element.content == nil ? .zero : .text(">") <> .hardline)
+    <> .text(">") <> (element.content == nil ? .empty : .hardline)
 }
 
 private func prettyPrintCloseTag(element: Element) -> Doc {
   return element.content == nil
-    ? .text(" />")
```
mbrandonw: turns out the valid way to do a "void" tag is just closing with `>` not `/>`, e.g. ``

Sources/HtmlPrettyPrint/Pretty.swift
```diff
@@ -81,6 +84,17 @@ private func prettyPrint(attribute: Attribute) -> Doc {
       <> .text("\"")
   }
 
+  if attribute.key == "content" {
```
mbrandonw: `content` tags should prob also flow to multiple lines since they can get long, e.g. meta tags use this

Tests/HtmlPrettyPrintTests/__Snapshots__/HtmlPrettyPrintTests.testPretty.txt
```diff
@@ -25,6 +25,6 @@
     </p>
     <img id="cat"
          class="cat"
-         src="cat.jpg" />
+         src="cat.jpg">
```
mbrandonw: snapshot tests are great!

Sources/HtmlPrettyPrint/Pretty.swift
stephencelis: Ah haha, yeah this is true for HTML5 but old XHTML habits die hard. I also would worry that there may be some gnarly exceptions, but you found that wasn't the case?

mbrandonw: Yeah seems legit. Ole passed this along to me https://www.w3.org/TR/2011/WD-html-markup-20110405/syntax.html#void-element

Tests/HtmlPrettyPrintTests/__Snapshots__/HtmlPrettyPrintTests.testPretty.txt
stephencelis: Wanna update it to use the new `fileExtension` option? :D

mbrandonw: oh yes yes!

PR #16, opened 2017-07-18
support svg tag

mbrandonw:
basic support for now, @amotion needs it

### DIFF
--- a/Sources/HTML/Elements.swift
+++ b/Sources/HTML/Elements.swift
@@ -226,6 +226,10 @@ public func style(_ css: String) -> Node {
   return node("style", [], [.text(EncodedString(css))])
 }
 
+public func svg(_ attribs: [Attribute], _ svg: String) -> Node {
+    return node("svg", attribs, [.text(EncodedString(svg))])
+}
+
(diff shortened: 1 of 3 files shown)


Sources/HTML/Elements.swift
```diff
@@ -226,6 +226,10 @@ public func style(_ css: String) -> Node {
   return node("style", [], [.text(EncodedString(css))])
 }
 
+public func svg(_ attribs: [Attribute], _ svg: String) -> Node {
+    return node("svg", attribs, [.text(EncodedString(svg))])
```
stephencelis: This (and `style`, above), are currently unsafe, right?

mbrandonw: purposely so tho! they are allowed to be unencoded in those tags in particular.

was thinking we could have an `internal` function `safeString` that is just the initializer of `EncodedString`

the cool thing still is that outside this framework no one is allowed to insert unencoded text. but this framework does have to enforce the rules and know what it is doing

mbrandonw:
gonna close this cause I have a diff PR coming...

PR #19, opened 2017-07-19
Use static members on Stylesheet and Values

stephencelis:
Big API change but works really nicely for:

  - autocomplete
  - namespacing (`floatLeft` -> `.left`)

### DIFF
--- a/Sources/Css/Background.swift
+++ b/Sources/Css/Background.swift
@@ -8,8 +8,10 @@ extension Background {
   }
 }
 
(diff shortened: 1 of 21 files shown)


stephencelis:
Could take this further by defining static colors on `Color`...but waiting till we figure out the `Auto`/`Inherit`/`None` sitch.

stephencelis:
(We also may wanna rename `Stylesheet` back to `Css` for nicer qualified naming again.)

PR #21, opened 2017-07-23
Color vals

stephencelis:
Gives color the ability to be `auto`, `inherit`, etc.


Sources/Css/Color.swift
```diff
@@ -1,157 +1,315 @@
...
+
+extension Color: _ExpressibleByColorLiteral {
+  public init(colorLiteralRed red: Float, green: Float, blue: Float, alpha: Float) {
+    self = .rgba(toUInt8(red), toUInt8(green), toUInt8(blue), alpha)
   }
+}
 
-  public private(set) var hue: Int {
+public extension Color {
+  public private(set) var red: UInt8? {
```
stephencelis: These getters/setters are now no-ops.

Sources/Css/Color.swift
```diff
@@ -1,157 +1,315 @@
...
+      let (h, s, _) = rgb2hsl(r, g, b)
+      let (red, green, blue) = hsl2rgb(h, s, l)
+      self = .rgba(red, green, blue, a)
+    }
   }
 }
 
 public func darken(_ by: Float) -> (Color) -> Color {
-  return over(\Color.lightness) <| { $0 * (1 - by) }
+  return over(\Color.lightness) <| { $0.map { $0 * (1 - by) } }
```
stephencelis: Need some prisms!

Tests/CssTests/FullStylesheetTests.swift
```diff
@@ -9,15 +9,15 @@ class FullStylesheetTests: XCTestCase {
   func testABigStyleSheet() {
     let css: Stylesheet =
       body % (
-        .background(red)
+        .background(Color.red)
```
stephencelis: Background color needs to be qualified now, though this speaks maybe more to the fact that the attribute accepting only `color` is `background-color`.

Tests/CssTests/BorderTests.swift
```diff
@@ -6,7 +6,7 @@ class BorderTests: XCTestCase {
   func testBorders() {
     let css = p % (
       .borderRadius(topLeft: 1, topRight: 2, bottomRight: 3, bottomLeft: 4)
-        <> .borderColor(top: red, right: blue, bottom: green, left: red)
+        <> .borderColor(top: .red, right: .blue, bottom: .lime, left: .red)
```
stephencelis: So it turns out `rgb(0, 255, 0)` in CSS is actually `lime`?

PR #26, opened 2017-07-25
Remove stylesheet statics

stephencelis:
Also undoing the `Stylesheet` statics. Was seeing bad build times and more type checking ambiguity. Can reinvestigate in the future.


stephencelis:
Gonna merge when green!

PR #28, opened 2017-07-29
README Updates

stephencelis:
Updating the README with some changes from the last coupla weeks.

### DIFF
--- a/Sources/Html/View.swift
+++ b/Sources/Html/View.swift
@@ -27,3 +27,13 @@ extension View where M == [Node] {
     self.call = const([node])
   }
 }
+
+extension View where A == (), M == [Node] {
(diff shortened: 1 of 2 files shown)


Sources/Html/View.swift
```diff
@@ -27,3 +27,13 @@ extension View where M == [Node] {
     self.call = const([node])
   }
 }
+
+extension View where A == (), M == [Node] {
```
stephencelis: P cool: these overloads let us omit the generic parameter from `View<()>` where we don't care about the data.

mbrandonw: nice nice, but I believvveeeee I ran into problems with these overloads :( we'll soon see!

maybe write some tests to cover some common use cases too...

PR #34, opened 2017-08-12
Some HttpPipeline clean up

mbrandonw:
some various clean up stuff, will comment...


Package.swift
```diff
@@ -15,6 +15,7 @@ let package = Package(
     .library(name: "HtmlPrettyPrint", targets: ["HtmlPrettyPrint"]),
     .library(name: "HttpPipeline", targets: ["HttpPipeline"]),
     .library(name: "HttpPipelineHtmlSupport", targets: ["HttpPipelineHtmlSupport"]),
+    .library(name: "HttpPipelineTestSupport", targets: ["HttpPipelineTestSupport"]),
```
mbrandonw: adds snapshot support to `Response`.

it would be fun to be able to snapshot middleware by providing some middleware and a connection and it prints out the transition states to the snapshot too.

Sources/HttpPipeline/Format.swift
```diff
@@ -1,19 +0,0 @@
-public enum Format {
```
mbrandonw: I got rid of this in favor of `MediaType`. it seems that `MediaType` includes more things than what is necessary for `Content-Type` header, but there is still a ton of overlap. think it's worth finding a way to use it for both?

Sources/HttpPipeline/Middleware.swift
```diff
@@ -1,61 +1,68 @@
...
+        status: status,
+        headers: conn.response.headers,
+        body: nil
       )
     )
   }
 }
 
+public func writeHeader<A>(_ header: ResponseHeader) -> Middleware<HeadersOpen, HeadersOpen, A, A> {
+  return \.response.headers %~ { $0 + [header] }
```
mbrandonw: if we had a free function for appending we could even do `\.response.headers %~ appended(header)` or something

Sources/HttpPipeline/Middleware.swift
```diff
@@ -1,61 +1,68 @@
...
 
-public func closeHeaders<A>() -> Middleware<HeadersOpen, BodyOpen, A, A> {
-  return { conn in
-    .init(
-      data: conn.data,
-      request: conn.request,
-      response: conn.response
-    )
-  }
+public func closeHeaders<A>(conn: Conn<HeadersOpen, A>) -> Conn<BodyOpen, A> {
```
mbrandonw: moved this to a function `Conn -> Conn` instead of `() -> Middleware` that way we can compose with `>>> closeHeaders` instead of `>>> closeHeaders()`.

Sources/HttpPipeline/Middleware.swift
```diff
@@ -1,61 +1,68 @@
...
-  { conn in
-    .init(
-      data: conn.data,
-      request: conn.request,
-      response: Response(
-        status: conn.response.status,
-        headers: conn.response.headers,
-        body: conn.data
-      )
+public func end(conn: Conn<BodyOpen, Data?>) -> Conn<ResponseEnded, Data?> {
```
mbrandonw: made `end` into a function so that there could be two overloads. this is the normal one...

Sources/HttpPipeline/Middleware.swift
```diff
@@ -1,61 +1,68 @@
...
+      headers: conn.response.headers,
+      body: conn.data
     )
+  )
 }
 
-public func redirect<A>(_ location: String) -> Middleware<StatusLineOpen, HeadersOpen, A, A> {
-  return writeStatus(.found)
-    >>> writeHeader("Location", location)
+public func end<A>(conn: Conn<HeadersOpen, A>) -> Conn<ResponseEnded, Data?> {
```
mbrandonw: and this one closes from `HeadersOpen` by providing a `nil` body

Sources/HttpPipeline/Middleware.swift
```diff
@@ -1,61 +1,68 @@
...
+    |> closeHeaders
+    |> map(const(nil)) >>> end
+}
+
+public func redirect<A>(
+  to location: String,
+  headersMiddleware: @escaping Middleware<HeadersOpen, HeadersOpen, A, A> = id
+  )
+  ->
+  Middleware<StatusLineOpen, ResponseEnded, A, Data?> {
```
mbrandonw: replaced `redirect` with this one which only allows setting some additional headers but not setting the body.

Sources/HttpPipeline/Middleware.swift
```diff
@@ -99,30 +106,24 @@ public func send(_ data: Data?) -> Middleware<BodyOpen, BodyOpen, Data?, Data?>
 }
 
 public func respond<A>(text: String) -> Middleware<HeadersOpen, ResponseEnded, A, Data?> {
-  return { conn in
-    conn.map(const(text.data(using: .utf8)))
-      |> writeHeader(.contentType(.text))
-      |> closeHeaders()
-      |> end
-  }
+  return map(const(text.data(using: .utf8)))
```
mbrandonw: was able to write some of helpers in point-free by pre-composing (i.e. `map` on functions) with a `map(const)` on `Conn`.

Sources/HttpPipeline/ResponseHeader.swift
```diff
@@ -1,6 +1,8 @@
+import MediaType
+
 public enum ResponseHeader {
   case allow([Method])
-  case contentType(Format)
+  case contentType(MediaType)
```
mbrandonw: it does seem like we can re-use all your work from `MediaType` right?

Sources/HttpPipelineHtmlSupport/Support.swift
```diff
@@ -8,7 +8,7 @@ public func respond<A>(_ view: View<A>) -> Middleware<HeadersOpen, ResponseEnded
     conn.map(view.rendered(with:))
       |> map { $0.data(using: .utf8) }
       |> writeHeader(.contentType(.html))
-      |> closeHeaders()
+      |> closeHeaders
```
mbrandonw: this is what the function change gave us

Sources/HttpPipelineTestSupport/HttpPipelineTestSupportTests.swift
```diff
@@ -0,0 +1,37 @@
...
+      .flatMap {
+        if case let .contentType(mediaType) = $0 {
+          return mediaType
+        }
+        return nil
+    }.first
+
+    if contentMediaType?.isApplication == .some(true) || contentMediaType?.isText == .some(true) {
+      // todo: use proper encoding when available
+      return top + "\n\n\(self.body.flatMap { String(data: $0, encoding: .utf8) } ?? "")\n"
```
mbrandonw: this is the snapshot test. it shows the status, headers, # of bytes, and then prints the body the body can be printed to a string it is...

Tests/HttpPipelineTests/HttpPipelineTests.swift
```diff
@@ -1,24 +1,74 @@
...
-//""",
-//                   pipeline(conn).response.description)
+    let middleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data?> =
+      writeStatus(.ok)
+        >>> respond(text: "Hello, world")
+
+    assertSnapshot(matching: middleware(conn).response)
+  }
+
+  func testHtmlResponse() {
```
mbrandonw: wrote a bunch of new tests

PR #36, opened 2017-08-15
Fix form data crash

mbrandonw:
found a crash while messing with some stuff

### DIFF
--- a/Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
+++ b/Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
@@ -95,7 +95,7 @@ class ApplicativeRouterTests: XCTestCase {
 
     let formString = "email="
     let formData = formString.data(using: .utf8)!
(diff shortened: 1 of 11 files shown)


Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
```diff
@@ -88,6 +88,24 @@ class ApplicativeRouterTests: XCTestCase {
     let formData = formString.data(using: .utf8)!
     XCTAssertEqual(.postInt(50), router.match(req(.post, "/post", formData)))
   }
+
+  func testMissingFormFieldValue() {
+    let router =
+      PostTestRoute.postString <¢> (.post *> .formField("email")) <* lit("post") <*| end
+
+    let formString = "email="
```
stephencelis: Wonder if this should change our call to `split` to use `omittingEmptySubsequences` so that the above would actually return an empty string for the value:

``` swift
$0.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
```

PR #40, opened 2017-08-18
Snapshot fix

mbrandonw:
What started as a snapshot fix for `Response` (it was using `snapshotFileExtension` instead of `snapshotPathExtension`) led me to making a few improvements and adding snapshotting to `Conn`, which is neat.


Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -7,15 +7,15 @@ import XCTest
...
+  public static var snapshotPathExtension: String? {
+    return "Response.txt"
   }
 
   public var snapshotFormat: String {
     let top = """
 Status \(self.status.rawValue) (\(self.status))
 Headers: [
-  \(self.headers.map { $0.description }.joined(separator: "\n  "))
+  \(self.headers.sorted(by: \.description).map { $0.description }.joined(separator: "\n  "))
```
mbrandonw: sorting headers, and using a new extension on array that can sort by key paths into comparables. this should be bridged with lenses in the prelude, but didn't wanna mess with that now

Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -35,3 +35,62 @@ Bytes: \(self.body?.count ?? 0)
     return top
   }
 }
+
+// todo: move to prelude
+extension Sequence {
+  fileprivate func sorted<C: Comparable>(by keyPath: KeyPath<Element, C>) -> [Element] {
```
mbrandonw: we can move this later...

Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -35,3 +35,62 @@ Bytes: \(self.body?.count ?? 0)
...
+    return self.sorted { lhs, rhs in
+      lhs[keyPath: keyPath] < rhs[keyPath: keyPath]
+    }
+  }
+}
+
+extension Conn: Snapshot {
```
mbrandonw: making `Conn` snapshottable so that we can capture the step and request

Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -35,3 +35,62 @@ Bytes: \(self.body?.count ?? 0)
...
+"""
+  }
+
+  public static var snapshotPathExtension: String? {
+    return "Conn.txt"
+  }
+}
+
+// Todo: move to snapshot lib
+extension URLRequest: Snapshot {
```
mbrandonw: this should be moved to the snapshot lib but can do some other time...

Tests/HttpPipelineTests/__Snapshots__/HttpPipelineTests/testBasicAuth_Authorized.1.Conn.txt
```diff
@@ -0,0 +1,22 @@
+Step:
```
mbrandonw: if we moved away from the `Snapshot` protocol and towards the struct version we could create snapshots of middleware where you provide a middleware and starting connection, and then we just apply the middleware and snapshot both connections. would be fun!

Tests/HttpPipelineTests/__Snapshots__/HttpPipelineTests/testCookies.1.Conn.txt
```diff
@@ -1,8 +1,23 @@
...
+
+Response:
+---------
 Status 200 (ok)
 Headers: [
-  Set-Cookie: user_id=123456
-  Set-Cookie: lang=es
   Content-Type: text/html; charset=utf8
+  Set-Cookie: lang=es
+  Set-Cookie: user_id=123456
```
mbrandonw: hm actually, are headers order dependent? what happens if you set the same cookie twice? does the last one override the first? 

if so,  maybe we shouldn't be sorting headers....

PR #44, opened 2017-08-24
Io middleware

mbrandonw:
I think we are gonna come across a lot of middleware that deals with `IO` values, so it would be nice to have a way to compose in that world. this is what I came up with. I named it `>->` cause it's kleisli like, yet I didn't use `flatMap` at all 🤔 

ill open a PR on pointfree to show how I used this. well, I haven't even used `>->` yet, but I think it'll be useful. I think you'll be able to compose middlewares like:

```swift 
fetchCurrentUser 
  >-> recordEpisodeView
  >>> resolve
  >>> writeStatus(.ok) ...
```


mbrandonw:
I deleted the io middleware stuff in favor of https://github.com/pointfreeco/swift-prelude/pull/64/. now this pr just contains some clean up stuff

PR #47, opened 2017-09-06
Viewport meta attributes

stephencelis:
Some viewport stuff.

``` swift
meta(viewport: .width(.deviceWidth), .initialScale(1))
// meta(viewport: .width(.deviceWidth) >| [.initialScale(1)])
```

### DIFF
--- a/Sources/Html/Elements.swift
+++ b/Sources/Html/Elements.swift
@@ -517,7 +517,7 @@ public func li<T: ContainsList>(_ content: [Node]) -> ChildOf<T> {
 }
 
 public func link(_ attribs: [Attribute<Element.Link>]) -> ChildOf<Element.Head> {
-  return .init(node("li", attribs, nil))
+  return .init(node("link", attribs, nil))
 }
 
 public func main(_ attribs: [Attribute<Element.Main>], _ content: [Node]) -> Node {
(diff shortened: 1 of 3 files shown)


Sources/Html/Elements.swift
```diff
@@ -516,7 +517,7 @@ public func li<T: ContainsList>(_ content: [Node]) -> ChildOf<T> {
 }
 
 public func link(_ attribs: [Attribute<Element.Link>]) -> ChildOf<Element.Head> {
-  return .init(node("li", attribs, nil))
+  return .init(node("link", attribs, nil))
```
stephencelis: Snuck in some fixes. This one was bad 😬

Sources/Html/Elements.swift
```diff
@@ -582,16 +583,89 @@ public func meta(keywords: [String]) -> ChildOf<Element.Head> {
...
+  public init(integerLiteral value: Int) {
+    self = .px(value)
+  }
+}
+
+public func meta(viewport props: NonEmptyArray<Viewport>) -> ChildOf<Element.Head> {
+  return meta([name(.viewport), content(Array(props).map(get(\.description)).joined(separator: ","))])
+}
+
+public func meta(viewport prop: Viewport, _ props: Viewport...) -> ChildOf<Element.Head> {
```
stephencelis: Been doing a thing where I overload functions that take `NonEmpty` with a friendlier forced variadic version. Not sure if it scales well, but it made `Gen` a bit more wieldy.

PR #50, opened 2017-09-11
Use string literals for the content of script and svg tags

mbrandonw:
This at least limits security vulnerabilities to the user of the lib putting in an unsafe literal string.

### DIFF
--- a/Sources/Html/Elements.swift
+++ b/Sources/Html/Elements.swift
@@ -789,7 +789,7 @@ public func samp(_ content: [Node]) -> Node {
   return samp([], content)
 }
 
-public func script(_ attribs: [Attribute<Element.Script>], _ content: StringLiteralType) -> Node {
+public func script(_ attribs: [Attribute<Element.Script>], _ content: StaticString) -> Node {
   return node("script", attribs, [.text(EncodedString(content))])
 }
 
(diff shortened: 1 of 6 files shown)


Sources/Html/Html.swift
```diff
@@ -59,19 +59,19 @@ public func document(_ content: [Node]) -> Node {
   return .document(content)
 }
 
-public func node<T>(_ name: String, _ attribs: [Attribute<T>], _ content: [Node]?) -> Node {
+public func node<T>(_ name: StringLiteralType, _ attribs: [Attribute<T>], _ content: [Node]?) -> Node {
```
mbrandonw: some of the stuff in this file can also be string literals

Sources/Html/Elements.swift
```diff
@@ -788,27 +789,27 @@ public func samp(_ content: [Node]) -> Node {
   return samp([], content)
 }
 
-public func script(_ attribs: [Attribute<Element.Script>], _ content: String) -> Node {
+public func script(_ attribs: [Attribute<Element.Script>], _ content: StaticString) -> Node {
```
mbrandonw: oops I misunderstood `StringLiteralType`, what I really wanted was `StaticString`

mbrandonw:
ok pushed a few more things. really this kinda became an audit of what `String`'s could be made into `StaticString`. I think there are a few more spots but this is a good start!

PR #52, opened 2017-09-12
Middleware transformer for adding content length

mbrandonw:


mbrandonw:
> Cool! The only thing here is that you have to manually inject it into the chain, and I think middleware that already knows it's rendering an entire response could do that for us. See #53.

yeah def, that's way better, but gonna merge this just so that we have more examples of middleware transformers in the lib...

Issue #43, opened 2017-08-23
different snapshot format for responses?


mbrandonw:
we can use the HTTP message format, e.g.:

```
curl -v 0.0.0.0:8080
* Rebuilt URL to: 0.0.0.0:8080/
*   Trying 0.0.0.0...
* TCP_NODELAY set
* Connected to 0.0.0.0 (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: 0.0.0.0:8080
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 24 Aug 2017 13:59:09 GMT
< Content-Type: text/html; charset=utf8
< Content-Length: 5721
< Connection: Keep-Alive
< Keep-Alive: timeout=60, max=99
< 
<!DOCTYPE html>```

stephencelis:
Fun thing to revisit in the future! Closing for now.

PR #53, opened 2017-09-12
Automatically include Content-Length

stephencelis:
Alternative (or supporting?) idea to https://github.com/pointfreeco/swift-web/pull/52.

### DIFF
--- a/Sources/HttpPipeline/Middleware.swift
+++ b/Sources/HttpPipeline/Middleware.swift
@@ -121,7 +121,7 @@ public func respond<A>(json: String) -> Middleware<HeadersOpen, ResponseEnded, A
   return respond(body: json, contentType: .json)
 }
 
-private func respond<A>(body: String, contentType: MediaType)
+public func respond<A>(body: String, contentType: MediaType)
   -> Middleware<HeadersOpen, ResponseEnded, A, Data?> {
     let data = body.data(using: .utf8)
     return map(const(data))
--- a/Sources/HttpPipeline/Middleware.swift
+++ b/Sources/HttpPipeline/Middleware.swift
@@ -98,7 +98,7 @@ public func send(_ data: Data?) -> Middleware<BodyOpen, BodyOpen, Data?, Data?> 
     }
 
     return .init(
-      data: conn.data,
+      data: concatenatedData,
       request: conn.request,
       response: Response(
         status: conn.response.status,
--- a/Sources/HttpPipeline/Middleware.swift
+++ b/Sources/HttpPipeline/Middleware.swift
@@ -1,4 +1,5 @@
 import Foundation
+import MediaType
 import Optics
 import Prelude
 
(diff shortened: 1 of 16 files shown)


Sources/HttpPipeline/Middleware.swift
```diff
@@ -109,24 +110,25 @@ public func send(_ data: Data?) -> Middleware<BodyOpen, BodyOpen, Data?, Data?>
...
 
 public func respond<A>(json: String) -> Middleware<HeadersOpen, ResponseEnded, A, Data?> {
-  return map(const(json.data(using: .utf8)))
-    >>> writeHeader(.contentType(.json))
-    >>> closeHeaders
-    >>> end
+  return respond(body: json, contentType: .json)
+}
+
+private func respond<A>(body: String, contentType: MediaType)
```
stephencelis: This should maybe be public?

Tests/HttpPipelineTests/__Snapshots__/HttpPipelineTests/testBasicAuth_Authorized.1.Conn.txt
```diff
@@ -13,10 +13,8 @@ Body: (Data, 0 bytes)
 
 Response:
 ---------
-Status 200 (ok)
-Headers: [
-  Content-Type: text/html; charset=utf8
-]
-Bytes: 19
+Status 200 OK
```
stephencelis: Having a little fun with the response formatting to make it look more like a raw HTTP response.

mbrandonw:
> Having a little fun with the response formatting to make it look more like a raw HTTP response.

oh yeah I have an issue here to format it like HTTP/1.1 message format:

https://github.com/pointfreeco/swift-web/issues/43

Sources/HttpPipeline/Middleware.swift
mbrandonw: yeah let's make it public!

Tests/HttpPipelineTests/HttpPipelineTests.swift
```diff
@@ -84,7 +84,11 @@ class HttpPipelineTests: XCTestCase {
     let middleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data?> =
       contentLength
         <| writeStatus(.ok)
-        >>> respond(html: "<p>Hello, world</p>")
+        >>> writeHeader(.contentType(.html))
+        >>> closeHeaders
+        >>> map(const(Data()))
```
stephencelis: `closeHeaders` should maybe map to an empty `Data` automatically?

Sources/HttpPipeline/Middleware.swift
```diff
@@ -97,7 +98,7 @@ public func send(_ data: Data?) -> Middleware<BodyOpen, BodyOpen, Data?, Data?>
     }
 
     return .init(
-      data: conn.data,
+      data: concatenatedData,
```
stephencelis: Failure to set this prevented `send` from working. Not sure if this is the right fix though? Feeling our data model may not be quite what we want it to be here.

stephencelis:
Gonna merge this for now, but definitely reveals that incremental middleware (anything that chunks a payload over time) needs some work.

PR #55, opened 2017-09-25
New middleware transformers

mbrandonw:
We had a few middleware transformers built in https://www.github.com/pointfreeco/pointfreeco that I moved over here. I also moved some transformers out of the `Middleware.swift` and into a dedicated shared transformers file.


Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -2,7 +2,41 @@ import Foundation
...
+///
+/// - Parameters:
+///   - user: The user name to authenticate against.
+///   - password: The password to authenticate against.
+///   - authenticateMiddleware: The middleware to run in the case that authentication fails.
+/// - Returns: Transformed middleware
+public func basicAuth<A>(
+  user: String,
+  password: String,
+  authenticateMiddleware: @escaping Middleware<HeadersOpen, ResponseEnded, A, Data?>
```
stephencelis: Whadya think of just calling this one `authenticate` or `middleware`?

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -16,7 +50,7 @@ public func basicAuth<A>(user: String, password: String)
           (
             writeStatus(.unauthorized)
               >>> writeHeader("WWW-Authenticate", "Basic")
```
stephencelis: Was reminded that you can configure a "realm" to customize what the pop-up says. _E.g._:

```
WWW-Authenticate: Basic realm="My App"
```

So maybe we need to add the enum now?

``` .swift
writeHeader(.wwwAuthenticate(.basic(realm: ""))
// vs.
writeHeader(.wwwAuthenticate(.basic(realm: "My App"))
```

mbrandonw: Ah yeah, nice one, will do!

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
mbrandonw: Yeah could do `authenticate`. Want something more specific than `middleware` cause it’s only run when there’s an auth failure. Could maybe even call it `failure`

stephencelis: yeah I like that!

mbrandonw:
just pushed this https://github.com/pointfreeco/swift-web/pull/55/commits/07f3e34682c06b4ef0e9cc10b9cc9d9fc1412bb7 that addresses the PR feedback

mbrandonw:
went ahead and added this https://github.com/pointfreeco/swift-web/pull/55/commits/608acb009e5b31c786d7f5b86b1a8992be16dc4c in order to fix printing status codes like `MOVED PERMANENTLY` as `MOVEDPERMANENTLY`

PR #56, opened 2017-09-26
Added routing middleware

mbrandonw:
One last piece of middleware transformer I forgot from my last PR #55. This takes a router, some not found middleware, and does the routing.

I added this to a package `ApplicativeRouterHttpPipelineSupport` that depends on each the router and pipeline.


mbrandonw:
@stephencelis gonna merge this cause it's just me copy-pasting from the point free repo, but lemme know if you have any questions/concerns

PR #57, opened 2017-09-27
Signed & Encrypted cookies

mbrandonw:
### What

Adds some helpers for creating signed and encrypted cookies. Don't know if there's a reason to add both, so up to discuss simplifying and allowing only encrypted. The difference is that signed cookies can be read by the user but not written, and encrypted can neither be read nor written. Encrypted cookies are also a bit more bytes.

To do this I needed some encryption. I decided to use IBM's [“BlueCryptor”](https://github.com/IBM-Swift/BlueCryptor), which just does a bunch of `#if` logic around using `CommonCrypto` for mac and `OpenSSL` for linux.

Kind of a bummer to have to bring this dependency into `swift-web`. Maybe it's ok for now, but in the future maybe we want the cookie functions to take a function as an argument that does the HMAC signing. Then we could have a supporting package that depends on `BlueCryptor` so that people can better opt out of taking in this dependency.

### How

There are 3 helpers for setting a signed/encrypted cookie: one for signing `Data`, one for `String` and another for any `Encodable`. Then there's a corresponding helper for each of those that “verifies” a signature and returns the data/string/decodable if it succeeds. The ability to encrypt is specified by a boolean flag in these methods.

I mostly followed the implementation from rails in which the cookie set has the form `"\(data)--\(digest)"`, where `digest` is computed from `data` and a secret via HMAC. And if the user wants to encrypt I just encrypt that whole thing.


Package.swift
```diff
@@ -46,8 +47,10 @@ let package = Package(
 
     .target(name: "HtmlTestSupport", dependencies: ["HtmlPrettyPrint", "SnapshotTesting"]),
 
-    .target(name: "HttpPipeline", dependencies: ["MediaType", "Prelude", "Optics"]),
-    .testTarget(name: "HttpPipelineTests", dependencies: ["HttpPipeline", "SnapshotTesting", "HttpPipelineTestSupport"]),
+    .target(name: "HttpPipeline",
+            dependencies: ["Cryptor", "MediaType", "Prelude", "Optics"]),
+    .testTarget(name: "HttpPipelineTests",
+                dependencies: ["Deriving", "HttpPipeline", "SnapshotTesting", "HttpPipelineTestSupport"]),
```
mbrandonw: I'm using `Deriving` in a test!

Sources/HttpPipeline/ResponseHeader.swift
```diff
@@ -1,14 +1,87 @@
 import MediaType
+import Prelude
 
 public enum ResponseHeader {
   case allow([Method])
   case contentLength(Int)
   case contentType(MediaType)
   case location(String)
-  case setCookie([String: String])
+  case setCookie(key: String, value: String, options: Set<CookieOption>)
```
mbrandonw: turns out it's not legit to set multiple cookies in one go, so had to change this up. also added support for cookie options like `HttpOnly`, `Max-Age`, etc...

Sources/HttpPipeline/SignedCookies.swift
```diff
@@ -0,0 +1,148 @@
...
+  /// Computes a signed cookie.
+  ///
+  /// - Parameters:
+  ///   - key: The name of the cookie.
+  ///   - value: The data of the cookie to sign.
+  ///   - options: (Optional) Extra options to attach to the cookie.
+  ///   - secret: The secret to sign the cookie with. This value must also be provided to verify a signature.
+  ///   - encrypt: (Optional) Further encrypts the signed cookie using the secret provided.
+  /// - Returns: A `Set-Cookie` header containing the signed cookie.
+  public static func setSignedCookie(
```
mbrandonw: these are static funcs on `ResponseHeader` so you can use it like you would `writeHeader(.setCookie(...))`

Sources/HttpPipeline/SignedCookies.swift
```diff
@@ -0,0 +1,148 @@
...
+  let secretBytes = CryptoUtils.byteArray(fromHex: secret)
+  let iv = [UInt8](repeating: 0, count: secret.count)
+  let encryptedTextBytes = CryptoUtils.byteArray(fromHex: encryptedText)
+
+  let decryptedText = Cryptor(operation: .decrypt, algorithm: .aes, options: .none, key: secretBytes, iv: iv)
+    .update(byteArray: encryptedTextBytes)?
+    .final()
+
+  return decryptedText
+    .map { Data($0.filter { $0 != 0 }) }
```
mbrandonw: gotta filter out `0` from the decrypted bytes since we padded in the `encrypted` function...

Tests/HttpPipelineTests/SignedCookieTests.swift
```diff
@@ -0,0 +1,172 @@
...
+      ResponseHeader.verifiedValue(
+        signedCookieValue: encryptedCookieValue,
+        secret: "deadbeefdeadbeefdeadbeefdead1234"
+      ),
+      "Reading signed cookie with wrong credentials returns nil."
+    )
+  }
+}
+
+struct Episode: Codable, DerivingEquatable {
```
mbrandonw: I was literally about to write my own `Equatable` instance like a chump

Sources/HttpPipeline/SignedCookies.swift
```diff
@@ -0,0 +1,148 @@
...
+  ///   - encrypt: (Optional) Further encrypts the signed cookie using the secret provided.
+  /// - Returns: A `Set-Cookie` header containing the signed cookie.
+  public static func setSignedCookie(
+    key: String,
+    data: Data,
+    options: Set<CookieOption> = [],
+    secret: String,
+    encrypt: Bool = false
+    )
+    -> ResponseHeader? {
```
mbrandonw: I had to make the decision of whether or not the return value should be optional. there are a lot of things that can go wrong along the way, and so it seemed weird to coalesce to a non-`nil` value at the end.

we may even want to use `Either` have a proper error enum

Sources/HttpPipeline/ResponseHeader.swift
```diff
@@ -23,11 +96,10 @@ public enum ResponseHeader {
...
     case let .location(uri):
       return ("Location", uri)
-    case let .setCookie(cookies):
-      let str = cookies.map { (pair: (key: String, value: String)) in
-        pair.key + "=" + pair.value // TODO: escape
-      }.joined(separator: "; ")
-      return ("Set-Cookie", str)
+    case let .setCookie(key: key, value: value, options: options):
+      // TODO: escape
+      let headerValue = (["\(key)=\(value)"] + options.map(get(\.description))).joined(separator: "; ")
```
mbrandonw: looks like cookie value escaping isn't so hard. it seems to only be spaces, commas, semicolons and a few control characters (\x00 to \x1F plus \x7F). we should tackle that soon

Sources/HttpPipeline/SignedCookies.swift
```diff
@@ -0,0 +1,149 @@
...
+
+      let encodedValue = data.base64EncodedString()
+      guard let computedDigest = digest(value: encodedValue, secret: secret) else { return nil }
+
+      let signedValue = "\(encodedValue)--\(computedDigest)"
+      guard let finalValue = encrypt ? encrypted(text: signedValue, secret: secret) : signedValue
+        else { return nil }
+
+      return .some(.setCookie(key: key, value: finalValue, options: options.union([.httpOnly])))
```
mbrandonw: @stephencelis so I originally had all signed/encrypted cookies be of type `HttpOnly` and `Secure`, but the latter one kinda complicates the local dev story. it seems like it's ok to transport this cookie over non-https since it's already encrypted, and I can't find any information online. whatdya think?

Sources/HttpPipeline/ResponseHeader.swift
```diff
@@ -1,14 +1,99 @@
...
+      }
+    }
+  }
+
+  public static func setCookies(_ pairs: [String: String], options: Set<CookieOption>) -> [ResponseHeader] {
+    return pairs.map { .setCookie(key: $0.key, value: $0.value, options: options) }
+  }
+
+  public static func clearCookie(key: String) -> ResponseHeader {
+    return .setCookie(key: key, value: "", options: [CookieOption.maxAge(0), .expires(0)])
```
mbrandonw: @stephencelis  I also don't know if this is the legit way to clear a cookie.

PR #58, opened 2017-10-01
[Proof of concept] Make middleware go `(Conn) -> IO<Conn>`

mbrandonw:
We've talked about this IRL, so I wanted to give it a shot. I changed `Middleware<I, J, A, B>` to be functions `(Conn<I, A>) -> IO<Conn<J, B>>` so that all middleware can do side effects in `IO`. This will help with composition in point free since right now we have a weird hodgepodge of middleware that returns pure `Conn`'s and `IO<Conn>`'s.

it does add some complexity tho! so wanna think it through. whatdya think?

### DIFF
--- a/Sources/HttpPipeline/Middleware.swift
+++ b/Sources/HttpPipeline/Middleware.swift
@@ -3,70 +3,81 @@ import MediaType
 import Optics
 import Prelude
 
-public typealias Middleware<I, J, A, B> = (Conn<I, A>) -> Conn<J, B>
+public typealias Middleware<I, J, A, B> = (Conn<I, A>) -> IO<Conn<J, B>>
 
 public func writeStatus<A>(_ status: Status) -> Middleware<StatusLineOpen, HeadersOpen, A, A> {
   return { conn in
-    .init(
-      data: conn.data,
-      request: conn.request,
-      response: conn.response |> \.status .~ status
+    pure(
+      .init(
+        data: conn.data,
+        request: conn.request,
+        response: conn.response |> \.status .~ status
+      )
     )
   }
 }
 
 public func writeHeader<A>(_ header: ResponseHeader) -> Middleware<HeadersOpen, HeadersOpen, A, A> {
-  return \.response.headers %~ { $0 + [header] }
+  return (\.response.headers %~ { $0 + [header] }) >>> pure
 }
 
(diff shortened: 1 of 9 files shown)


Sources/HttpPipeline/Middleware.swift
```diff
@@ -47,26 +47,26 @@ public func end(conn: Conn<BodyOpen, Data?>) -> Conn<ResponseEnded, Data?> {
   )
 }
 
-public func end<A>(conn: Conn<HeadersOpen, A>) -> Conn<ResponseEnded, Data?> {
+public func end<A>(conn: Conn<HeadersOpen, A>) -> IO<Conn<ResponseEnded, Data?>> {
   return conn
     |> closeHeaders
-    |> map(const(nil))
-    |> end
+    >-> map(const(nil)) >>> pure
```
mbrandonw: middlewares now compose with kleisli `>->`, but middleware transformers will still compose with regular `<<<`

mbrandonw: oh and anything that is pure (like the `map(const(nil))` above) needs to be post composed with a `>>> pure`.

Tests/ApplicativeRouterHttpPipelineSupportTests/ApplicativeRouterHttpPipelineSupportTests.swift
```diff
@@ -14,23 +14,21 @@ class ApplicativeRouterHttpPipelineSupportTests: XCTestCase {
...
-        <| (
-          writeStatus(.ok)
-            >>> { $0 |> respond(text: "Recognized route: \($0.data)") }
-    )
+        <| writeStatus(.ok)
+        >-> { $0 |> respond(text: "Recognized route: \($0.data)") }
 
     assertSnapshot(
-      matching: middleware(connection(from: URLRequest(url: URL(string: "/")!))),
+      matching: middleware(connection(from: URLRequest(url: URL(string: "/")!))).perform(),
```
mbrandonw: I wish I didn't need to do `.perform()` and instead could do `>>> perform` at the end of the middleware declaration above, _BUT_ swift gets confused with the `.perform` on `NSObject` since we are in an `XCTest` :/

mbrandonw:
(LGTM. —stephen)

PR #59, opened 2017-10-07
Added a bunch of tests for router.

mbrandonw:
### What

I added a bunch of tests to the applicative router to show more uses of it. I mostly did this to get a better understanding of how to construct complicated routers. There are a few fun things in here.

### DIFF
--- a/Sources/ApplicativeRouter/ApplicativeRouter.swift
+++ b/Sources/ApplicativeRouter/ApplicativeRouter.swift
@@ -167,7 +167,7 @@ extension Parser {
   }
 
   public static var int: Parser<I, Int> {
-    return component { Int($0) }
+    return component(Int.init)
   }
 
   public static var str: Parser<I, String> {
--- a/Sources/ApplicativeRouter/ApplicativeRouter.swift
+++ b/Sources/ApplicativeRouter/ApplicativeRouter.swift
@@ -52,22 +52,6 @@ extension Parser {
   }
 }
 
-// MARK: - Monad
-
-extension Parser {
-  public func flatMap<B>( _ f: @escaping (A) -> Parser<I, B>) -> Parser<I, B> {
-    return Parser<I, B>(
-      parse: { route in
-        guard case let .some(rest: rest, match: match) = self.parse(route) else { return nil }
-        return f(match).parse(rest)
-    })
-  }
-
-  public static func >>- <B>(lhs: Parser, f: @escaping (A) -> Parser<I, B>) -> Parser<I, B> {
-    return lhs.flatMap(f)
-  }
-}
-
 // MARK: - Apply
 
 extension Parser {
--- a/Sources/ApplicativeRouter/ApplicativeRouter.swift
+++ b/Sources/ApplicativeRouter/ApplicativeRouter.swift
@@ -52,6 +52,22 @@ extension Parser {
   }
 }
 
+// MARK: - Monad
+
+extension Parser {
+  public func flatMap<B>( _ f: @escaping (A) -> Parser<I, B>) -> Parser<I, B> {
+    return Parser<I, B>(
+      parse: { route in
+        guard case let .some(rest: rest, match: match) = self.parse(route) else { return nil }
+        return f(match).parse(rest)
+    })
+  }
+
+  public static func >>- <B>(lhs: Parser, f: @escaping (A) -> Parser<I, B>) -> Parser<I, B> {
+    return lhs.flatMap(f)
+  }
+}
+
 // MARK: - Apply
 
 extension Parser {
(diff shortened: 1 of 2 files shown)


Sources/ApplicativeRouter/ApplicativeRouter.swift
```diff
@@ -52,6 +52,22 @@ extension Parser {
   }
 }
 
+// MARK: - Monad
+
+extension Parser {
+  public func flatMap<B>( _ f: @escaping (A) -> Parser<I, B>) -> Parser<I, B> {
```
mbrandonw: one of the things I wanted to explore needed `flatMap` for parsers, so adding it. will explain more below...

Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
```diff
@@ -110,24 +146,26 @@ class ApplicativeRouterTests: XCTestCase {
 
 enum Route {
   case home
-  case episodes
-  case episode(String)
+  case episodes(RefTag<Int?>)
```
mbrandonw: I wanted to experiment with how to derive more complicated parsers from simpler parsers. 

so, imagine you had a route that plucked out a bunch of data, and then in addition to that we want to pluck out a `ref` query param. we could include `ref: String?` in every route's associated value, _oooor_ we could create a wrapper type `RefTag<A> = (String?, a)` and allow _any_ route to be lifted to understand ref tags. That's what this route (and its tests) attempts to demonstrate.

This ref tag example is a total real life use case, tho I think instead of having `RefTag<A>` for every route we would flip it and have the router ultimately be a parser of `RefTag<Route>`. And really that might be generalized to a type `CommonData<A>` that wrapped `Route` and tried plucking out all values we expect to be common in all requests.

Fun!

Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
```diff
@@ -163,3 +201,25 @@ enum PostTestRoute: Equatable {
...
+/// A struct that holds a ref tag that can be plucked from a query string, along with whatever else
+/// could get plucked from the request.
+struct RefTag<A> {
+  let ref: String?
+  let rest: A
+}
+
+/// Converts a parser of `A` into a parser of `RefTag<A>` by first trying to parse out the `A`, and then
+/// parsing a query param `ref` and storing both values in `RefTag<A>`.
+func refTag<I, A>(_ parser: Parser<I, A>) -> Parser<I, RefTag<A>> {
```
mbrandonw: here I use `flatMap` to take any parsers of `A` and turn it into a parser of `RefTag<A>`. it's pretty fun!

also cool to see that `I` is allowed to be free since query params can exist on requests of all methods. however, if we tried this trick on trying to always attempt plucking out some post body data we could restrict it to only `POST`s (did anyone say _csrf tokens_?!). I'm really starting to dig the phantom param for http method.

Tests/ApplicativeRouterTests/ApplicativeRouterTests.swift
```diff
@@ -12,21 +12,57 @@ func req(_ method: ApplicativeRouter.Method, _ location: String, _ body: Data? =
...
-        <|> Route.signup <¢> (.post *> .dataBody) <* lit("signup") <*| end
+        // Route with path argument and two optional query params: /episode/:id?ref=foo&t=123
+        <|> curry(Route.episode) <¢> (.get <* lit("episode") *> .str) <*> opt(param("ref")) <*> opt(param("t", .int)) <*| end
+
+        // Route with no path arguments and two query params, `page` and `ref`. However, the `ref` param
+        // is parsed using the `refTag` helper that will provide a first class wrapper type for that value.
+        <|> Route.episodes <¢> refTag(.get <* lit("episodes") *> opt(param("page", .int))) <*| end
+
+        // Route with one query param: /search?query=foo
+        <|> Route.search <¢ .get <* lit("search") <*> opt(param("query")) <*| end
```
mbrandonw: I'm a lil confused why this is the correct way to express a route with a single query param. before, it was written as:

```swift 
Route.search <¢> (.get <* lit("search") *> opt(param("query"))) <*| end
```

and although that worked, it seemed counter-intuitive to have the `param` parser inside the parens where we do `.get <* lit` (lol). I wanted to bring it out of the parens so we could think of first parsing the path, and then the query params.

in order to do that, and get it in the form I now have, I had to change `<¢>` to `<¢`... can you help me understand that?

mbrandonw: oh nvm, I now get it. `Route.search <¢ .get` is executed first and we don't care about the result of the `.get` parser. nice!

mbrandonw:
lol will close cause this is all going away now :(

PR #61, opened 2017-10-18
Applicative parser-printer

mbrandonw:
### What

Re-implements the applicative router using the ideas from this [paper](http://www.informatik.uni-marburg.de/~rendel/unparse/).

The basic idea is to combine the concepts of a parser `(Request) -> A?` and printer `(A) -> Request?` into one applicative structure. A few complications arise when you try to do this, but it can all be dealt with and yer left with a router that can also generate urls from values.

### How

To get a single applicative structure on the parser and printer you gotta tweak a few things. First, the generic `A` appears on both the left and right side of the function arrow `->`, which means it needs to deal with covariant and contravariant things. To fix this we change our category so that morphisms consists of partial, invertible functions, i.e. a morphism from `A` to be `B` in this category is a pair of partial functions `(A) -> B?` + `(B) -> A?` that are inverses on the subset of values they are defined. We call this type `PartialIso<A, B>`, whereas the paper just calls it `Iso`.

With that set up, you basically redo all of the applicative parsing stuff using `PartialIso<A, B>` as arrows instead of functions, and it all just works. One caveat is that you can't use the standard applicative, you gotta use the one that is covariant+contravariant friendly, which has signature `f a -> f b -> f (a, b)`. But, then it's right associative instead of left. I'll add more notes about that in the PR review...

### Features

Cool thing is that we can now route + print + more all at the same time. For example, in the tests I have a router like:

```swift 
let router = 
  Routes.iso.episodeComment
    <¢> get %> lit("episodes") %> pathParam(.intOrString) 
    <%> lit("comments") %> .int
```

for matching urls of the form `/episodes/:param/comments/:id`. We can match the url like so:

```swift 
router.match(string: "episodes/intro-to-functions/comments/42")
// => Routes.episodeComment("intro-to-functions", 42)
```

then we can generate a URL to a route like so:

```swift 
router.url(for: .episodeComment("intro-to-functions", 42))
// => "episodes/intro-to-functions/comments/42"
```

and finally, we can generate a type-annotated template/documentation string from a route:

```swift
router.templateUrl(for: .episodeComment("", 0))
// => "episodes/:string_or_int/comments/:int"
```

Pretty cool!

### Todo

- [ ] Get the phantom type in place for handling GET/POST/etc in a type safe manner
- [ ] Perhaps do a full blown state machine in phantom types like we did with middleware
- [x] rename `image`/`preimage` to `apply`/`unapply`
- [ ] Can `PartialIso` be replaced by `Prism`?
- [ ] look into how [this](https://github.com/MedeaMelana/JsonGrammar2) uses tuples


Package.swift
```diff
@@ -27,8 +27,8 @@ let package = Package(
     .package(url: "https://github.com/IBM-Swift/BlueCryptor.git", .exact("0.8.17"))
   ],
   targets: [
-    .target(name: "ApplicativeRouter", dependencies: ["Either", "Prelude"]),
-    .testTarget(name: "ApplicativeRouterTests", dependencies: ["ApplicativeRouter", "Optics"]),
+    .target(name: "ApplicativeRouter", dependencies: ["Either", "Optics", "Prelude"]),
```
mbrandonw: I could do without `Optics` in the router if we didn't wanna bring in this dependency. just hard to say no to optics!

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -0,0 +1,190 @@
+import Foundation
+import Prelude
+import Optics
+
+/// Processes and consumes a single path component matching the string provided.
+public func lit(_ str: String) -> Router<Prelude.Unit> {
```
mbrandonw: I put all the request routing specific combinators in its own file so it's easy to see what is core printer-parser versus routing specific.

mbrandonw: also almost all of these combinators are straight forward ports from what we had previously.

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -0,0 +1,190 @@
...
+public let patch = method(.patch)
+public let post = method(.post)
+public let put = method(.put)
+
+// MARK: - Experimental
+
+// TODO: open for discussion
+infix operator <=>: infixr9
+
+public func <=> <A> (lhs: String, rhs: PartialIso<String, A>) -> Router<A> {
```
mbrandonw: im not sure how deep into custom operators we wanna go for routing. for example this allows us to do `"ref" <=> opt(.string)` instead of `pathParam("ref", opt(.string))`, etc. we could also have things like `</>`, `<?>`, `<&>`, and if we dod the state machine phantom types we'd be able to make sure the operators only work where we expect.

Sources/ApplicativeRouter/Method.swift
```diff
@@ -0,0 +1,10 @@
+// TODO: should we move this to its own package so both the router and pipeline can use?
```
mbrandonw: we should prob do this ^

Sources/ApplicativeRouter/PartialIso.swift
```diff
@@ -0,0 +1,98 @@
...
+// todo: use a profunctor Iso?
+//typealias Iso_<S, T, A, B> = ((S) -> A) -> ((B) -> T)
+
+// todo: move to prelude?
+
+// TODO: Change PartialIso to be Partial<A, B, E: Semigroup>: `((A) -> Validation<E, B>, (B) -> Validation<E, A>)
+
+/// A partial isomorphism is a partial function that is invertible on the parts it is defined, i.e.
+/// `image(x) = y` if and only `preimage(y) = x`.
+public struct PartialIso<A, B> {
```
mbrandonw: this is the core of what `PartialIso` is, and feel we should move to its own package in Prelude

Sources/ApplicativeRouter/RequestData.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+import Prelude
+
+struct RequestData: Monoid {
```
mbrandonw: this is just the value we parse from and print into. the parser-printer works over any monoid.

Sources/ApplicativeRouter/SomePartialIsos.swift
```diff
@@ -0,0 +1,140 @@
+import Either
```
mbrandonw: this file contains a bunch of helpful partial isomorphisms

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
+import Foundation
+import Prelude
+import Optics
+
+// MARK: - Syntax Router
+
+// TODO: should this be generic over any monoid `M` instead of using `RequestData` directly?
+// TODO: generic over an error semigroup too?
+public struct Router<A> {
```
mbrandonw: and now we finally get to the router!

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
...
+  let parse: (RequestData) -> (rest: RequestData, match: A)?
+  let print: (A) -> RequestData?
+  let template: (A) -> RequestData?
```
mbrandonw: the router simultaneously holds the functions that will parse, print and document routes.

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
...
+
+extension Router {
+  /// TODO: use `Tuple`?
+  /// Processes with the left and right side routers, and if they succeed returns the pair of their results.
+  public static func <%> <B> (lhs: Router, rhs: Router<B>) -> Router<(A, B)> {
+    return Router<(A, B)>(
+      parse: { str in
+        guard let (more, a) = lhs.parse(str) else { return nil }
+        guard let (rest, b) = rhs.parse(more) else { return nil }
+        return (rest, (a, b))
```
mbrandonw: applicative parsing is basically the same

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
...
+      },
+      print: { ab in
+        let lhsPrint = lhs.print(ab.0)
+        let rhsPrint = rhs.print(ab.1)
+        return (curry(<>) <¢> lhsPrint <*> rhsPrint) ?? lhsPrint ?? rhsPrint
```
mbrandonw: applicative printing using the monoid on `RequestData`

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
...
+    )
+  }
+}
+
+// Apply
+
+extension Router {
+  /// TODO: use `Tuple`?
+  /// Processes with the left and right side routers, and if they succeed returns the pair of their results.
+  public static func <%> <B> (lhs: Router, rhs: Router<B>) -> Router<(A, B)> {
```
mbrandonw: this is the alternate applicative. I decided to use `<%>` to represent it because 1) it's different from `<*>` in that it has a different signature and is right associative instead of left, and 2) although `<*>` induces a `<%>` the converse is not true (and in fact this router is such a counterexample).

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -0,0 +1,159 @@
...
+  /// Processes with the left and right side routers, discarding the result of the left side.
+  public static func %> (x: Router<Prelude.Unit>, y: Router) -> Router {
+    return (PartialIso.commute >>> PartialIso.unit.inverted) <¢> x <%> y
+  }
+}
+
+extension Router where A == Prelude.Unit {
+  /// Processes with the left and right routers, discarding the result of the right side.
+  public static func <% <B>(x: Router<B>, y: Router) -> Router<B> {
+    return PartialIso.unit.inverted <¢> x <%> y
```
mbrandonw: the `<%` and `%>` operators are precisely like `<*` and `*>` except they can _only_ work when the data being discarded is `Unit`. that's kinda neat

Sources/ApplicativeRouter/Todo.swift
```diff
@@ -0,0 +1,11 @@
...
+
+// todo: move to prelude: right associative applicative
+infix operator <%>: infixr4
+infix operator %>: infixr4
+infix operator <%: infixr4
+
+// todo: move to prelude
+public func >-> <A, B, C>(lhs: @escaping (A) -> B?, rhs: @escaping (B) -> C?) -> (A) -> C? {
+  return lhs >>> flatMap(rhs)
+}
```
mbrandonw: we should move everything in this file to prelude

Tests/ApplicativeRouterHttpPipelineSupportTests/ApplicativeRouterHttpPipelineSupportTests.swift
```diff
@@ -36,7 +36,7 @@ class ApplicativeRouterHttpPipelineSupportTests: XCTestCase {
   }
 
   func testRoute_UnrecognizedWithCustomNotFound() {
-    let router = Route.home <¢ .get <*| end
+    let router = Route.iso.home <¢> get <% _end
```
mbrandonw: updating this test shows how similar the new router is to the old. the `iso` stuff is explained below...

Tests/ApplicativeRouterHttpPipelineSupportTests/ApplicativeRouterHttpPipelineSupportTests.swift
```diff
@@ -55,4 +55,20 @@ class ApplicativeRouterHttpPipelineSupportTests: XCTestCase {
 enum Route {
   case home
   case episode(String)
+
+  enum iso {
+    static let home = parenthesize <| PartialIso<Prelude.Unit, Route>(
```
mbrandonw: the only boilerplate we incur with the new routes is that each enum case of our routing enum needs to have an partial isomorphism `PartialIso<(Tuple, Of, Args), Route>`. the `parenthesize` helper is kind like how we need `curry` in the other applicative router, except that it takes a flat list of tuple arguments and right-parenthesizes them so that we can use it with `<%>`

Tests/ApplicativeRouterTests/TestRouter.swift
```diff
@@ -0,0 +1,110 @@
...
+
+enum Routes {
+  case root
+  case pathComponents(param: Either<String, Int>, commentId: Int)
+  case postBodyField(email: String)
+  case postBodyJsonDecodable(episode: Episode, param: Int)
+  case simpleQueryParams(ref: String?, active: Bool, t: Int)
+}
+
+let router: Router<Routes> = [
```
mbrandonw: and this is a fun router with all types of interesting routes

Sources/ApplicativeRouter/Todo.swift
```diff
@@ -1,11 +0,0 @@
-import Prelude
```
mbrandonw: 👍 👍

Sources/ApplicativeRouter/SomePartialIsos.swift
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// TODO: possible to document this in the applicative of Syntax?
+
+extension PartialIso where A == String, B == Bool {
+  /// An isomorphism between strings and booleans.
+  public static var bool: PartialIso {
+    return .init(
+      apply: {
+        $0 == "true" || $0 == "1" ? true
```
ilyapuchka: I guess you can just do `apply: { $0 == "1" ? true : $0 == "0" ? false : Bool($0.lowercased()) }`

mbrandonw: @ilyapuchka  oh yeah, I think that would be equivalent! forgot that `Bool` has a constructor taking a string...

Tests/ApplicativeRouterTests/TestRouter.swift
```diff
@@ -0,0 +1,108 @@
...
+  // POST /signup
+  Routes.iso.postBodyField
+    <¢> post %> formField("email") <% lit("signup") <% end,
+
+  // POST /episodes/:id
+  Routes.iso.postBodyJsonDecodable
+    <¢> post %> jsonBody(Episode.self) <%> lit("episodes") %> .int <% end,
+
+  // /path/to/somewhere/cool?ref=:optional_string&active=:bool&t=:int
+  Routes.iso.simpleQueryParams
```
ilyapuchka: do you experience any compile time issues with long routes? for me swift compiler fails to compile more than five patterns in a row, i.e `"a" / "b" / "c" / "d" / "e" / "f"`. I wonder if it might be because I use a lot of variants of the same operator (`/`) with different generic types.

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -0,0 +1,185 @@
+import Foundation
+import Prelude
+import Optics
+
+/// Processes and consumes a single path component matching the string provided.
+public func lit(_ str: String) -> Router<Prelude.Unit> {
```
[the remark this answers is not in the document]
stephencelis: The old applicative router had `any` and `many` combinators. We should be able to add them back!

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -0,0 +1,185 @@
...
+  },
+    print: { a in
+      .init(method: nil, path: [f.unapply(a) ?? ""], query: [:], body: nil)
+  },
+    template: { a in
+      .init(method: nil, path: [":\(typeKey(a))"], query: [:], body: nil)
+  })
+}
+
+/// Processes (and does not consume) a query param keyed by `key` as a string.
```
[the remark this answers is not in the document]
mbrandonw: It could probably go either way, but one reason is that query params are not sequential whereas path components are.

PR #66, opened 2017-10-24
Request logger middleware transformer

mbrandonw:
This wraps middleware to log information about the requests that come through and time the middleware.


Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift
```diff
@@ -166,4 +166,19 @@ class SharedMiddlewareTransformersTests: XCTestCase {
...
   }
+
+  func testRequestLogger() {
+    var log: [String] = []
+    let middleware = requestLogger(logger: { log.append($0) })
+      <| writeStatus(.ok)
+      >-> writeHeader(.contentType(.html))
+      >-> closeHeaders
+      >-> map(const(Data())) >>> pure
+      >-> send("<p>Hello, world</p>".data(using: .utf8))
```
stephencelis: Think this needs to be rebased? `send` takes non-optional `Data` now.

PR #68, opened 2017-10-28
Improved data handling

stephencelis:
- Prefer typed guarantees of `String(decoding:as:)` over `String(data:encoding:)`
  - Prefer typed guarantees of `Data(string.utf8)` over `string.data(encoding: .utf8)`.
  - Rename `Conn<Step, Data>` to `Conn<Step, A>`
  - Don't use `Data?` in pipeline: respond with `Conn<ResponseEnded, Data>`, not `Conn<ResponseEnded, Data?>`

I'm not sure if we were depending on a semantic difference with `Data?`, but it seems unnecessary.

### DIFF
--- a/Sources/ApplicativeRouterHttpPipelineSupport/ApplicativeRouterHttpPipelineSupport.swift
+++ b/Sources/ApplicativeRouterHttpPipelineSupport/ApplicativeRouterHttpPipelineSupport.swift
@@ -11,10 +11,10 @@ import Prelude
 ///   - notFound: Middleware to run if a route is not recognized.
 public func route<A, Route>(
   router: Router<Route>,
-  notFound: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data?> = notFound(respond(text: "don't know that url"))
+  notFound: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data> = notFound(respond(text: "Not Found"))
   )
-  -> (@escaping Middleware<StatusLineOpen, ResponseEnded, Route, Data?>)
-  -> Middleware<StatusLineOpen, ResponseEnded, A, Data?> {
+  -> (@escaping Middleware<StatusLineOpen, ResponseEnded, Route, Data>)
+  -> Middleware<StatusLineOpen, ResponseEnded, A, Data> {
 
     return { middleware in
       return { conn in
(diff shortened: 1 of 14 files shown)


Sources/HttpPipeline/Conn.swift
```diff
@@ -6,46 +6,52 @@ public enum HeadersOpen {}
 public enum BodyOpen {}
 public enum ResponseEnded {}
 
-public struct Conn<Step, Data> {
-  public private(set) var data: Data
+public struct Conn<Step, A> {
```
mbrandonw: yeah this is way better. polymorphism for the win!

Sources/ApplicativeRouterHttpPipelineSupport/ApplicativeRouterHttpPipelineSupport.swift
```diff
@@ -11,10 +11,10 @@ import Prelude
 ///   - notFound: Middleware to run if a route is not recognized.
 public func route<A, Route>(
   router: Router<Route>,
-  notFound: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data?> = notFound(respond(text: "don't know that url"))
+  notFound: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data> = notFound(respond(text: "Not Found"))
```
mbrandonw: I _suppose_ that's better :/

Sources/HttpPipeline/Middleware.swift
```diff
@@ -86,24 +85,26 @@ public func send(_ data: Data?) -> Middleware<BodyOpen, BodyOpen, Data?, Data?>
...
-  -> Middleware<HeadersOpen, ResponseEnded, A, Data?> {
-    let data = body.data(using: .utf8)
+  -> Middleware<HeadersOpen, ResponseEnded, A, Data> {
+
+    let data = Data(body.utf8)
+
     return map(const(data)) >>> pure
       >-> writeHeader(.contentType(contentType))
-      >-> writeHeader(.contentLength(data?.count ?? 0))
+      >-> writeHeader(.contentLength(data.count))
```
mbrandonw: is this equivalent to `body.utf8.count`?

mbrandonw: never mind I didnt see that `data` was being used elsewhere. it thought we could avoid the `Data` creation...

PR #70, opened 2017-11-10
Basic plaintext from html

mbrandonw:
This is super basic, but just wanted to get something in place that we can improve later. I need this for some of the email work im doing.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,7 +24,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("9161df1")),
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("b4081b7")),
     .package(url: "https://github.com/bkase/DoctorPretty.git", .exact("0.3.0")),
-    .package(url: "https://github.com/IBM-Swift/BlueCryptor.git", .exact("0.8.17"))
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -24,7 +24,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("9161df1")),
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("b4081b7")),
     .package(url: "https://github.com/bkase/DoctorPretty.git", .exact("0.3.0")),
-    .package(url: "https://github.com/IBM-Swift/BlueCryptor.git", .exact("0.8.17"))
+    .package(url: "https://github.com/IBM-Swift/BlueCryptor.git", .exact("0.8.21"))
```
mbrandonw: unrelated to this PR, but seems the version of `CommonCrypto` that `BlueCryptor` depended on no longer exists, and so updating `BlueCryptor` to fix that

Sources/Html/PlainText.swift
```diff
@@ -0,0 +1,36 @@
+/// Naive transformation of HTML nodes into plain text.
+public func plainText(for nodes: [Node]) -> String {
+  return nodes.map(plainText(for:)).joined()
+}
+
+/// Naive transformation of an HTML into plain text.
```
stephencelis: Fun!

PR #72, opened 2017-11-10
Conn Error Handling

stephencelis:
I think https://github.com/pointfreeco/pointfreeco/pull/29 showed how complicated error handling composition was, so this PR attempts to bring error types into our connection and middleware. It's showing promise so far! But plenty of unanswered questions and we'll need to play around with actual errors a bit. Still WIP but wanted to document some discussion.


Sources/HttpPipeline/Middleware.swift
```diff
@@ -15,90 +15,86 @@ public func writeStatus<A>(_ status: Status) -> Middleware<StatusLineOpen, Heade
...
 
-public func end(conn: Conn<BodyOpen, Data>) -> IO<Conn<ResponseEnded, Data>> {
-  return pure <| .init(
-    data: conn.data,
-    request: conn.request,
-    response: Response(
-      status: conn.response.status,
-      headers: conn.response.headers,
-      body: conn.data
+public func end<E>(conn: Conn<BodyOpen, E, Data>) -> IO<Conn<ResponseEnded, E, Data>> {
```
stephencelis: This gets a lil complicated. There's definitely ways to clean up with some functions, but wonder if we can clean it up as is?

Sources/HttpPipeline/Middleware.swift
```diff
@@ -15,90 +15,85 @@ public func writeStatus<A>(_ status: Status) -> Middleware<StatusLineOpen, Heade
...
+        headers: conn.response.headers,
+        body: data
+      )
+    )
+  }
 }
 
-// TODO: rename to ignoreBody
-public func end<A>(conn: Conn<HeadersOpen, A>) -> IO<Conn<ResponseEnded, Data>> {
+public func ignoreBody<E, A>(conn: Conn<HeadersOpen, E, A>) -> IO<Conn<ResponseEnded, E, Data>> {
```
stephencelis: Makes things slightly easier to troubleshoot.

Sources/HttpPipeline/Middleware.swift
```diff
@@ -15,90 +15,85 @@ public func writeStatus<A>(_ status: Status) -> Middleware<StatusLineOpen, Heade
...
-      request: conn.request,
-      response: Response(
-        status: conn.response.status,
-        headers: conn.response.headers,
-        body: concatenatedData
-      )
-    )
-  }
+public func send<E>(_ data: Data) -> Middleware<BodyOpen, BodyOpen, E, E, Data, Data> {
+  return pure <<< map { $0 + data }
```
stephencelis: This one's actually pretty simple, though it would need an overhaul if we ever add some kind of continuation support for streaming endpoints.

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -164,9 +164,9 @@ private func makeHttps(url: URL) -> URL? {
 /// Transforms middleware into one that logs the request info that comes through and logs the amount of
 /// time the request took.
 public func requestLogger(
-  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data>
+  _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, Never, Never, Prelude.Unit, Data>
```
stephencelis: This is our only middleware with a concrete type. Kind of neat, though, to see that a logging middleware probably only makes sense if it captures _all_ responses.

Tests/ApplicativeRouterHttpPipelineSupportTests/ApplicativeRouterHttpPipelineSupportTests.swift
```diff
@@ -12,10 +12,17 @@ class ApplicativeRouterHttpPipelineSupportTests: XCTestCase {
       Route.iso.home <¢> get <% end
         <|> Route.iso.episode <¢> get %> lit("episode") %> .string <% end
 
-    let middleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
+    let middleware: Middleware<StatusLineOpen, ResponseEnded, Never, Never, Prelude.Unit, Data> =
       route(router: router)
         <| writeStatus(.ok)
-        >-> { $0 |> respond(text: "Recognized route: \($0.data)") }
+
+        >>> map(map { Data("Recognized route: \($0)".utf8) })
```
stephencelis: This stuff gets a little gross. There's certainly a composition I'm missing here, or maybe `ExceptT`-like transformer functions would help.

Tests/ApplicativeRouterHttpPipelineSupportTests/ApplicativeRouterHttpPipelineSupportTests.swift
```diff
@@ -12,10 +12,17 @@ class ApplicativeRouterHttpPipelineSupportTests: XCTestCase {
...
+        // FIXME: for snapshot to render body
+        >-> writeHeader(.contentType(.plain))
+        >-> { $0 |> writeHeader(.contentLength($0.data.right!.count)) }
```
stephencelis: These two lines are here because `end` doesn't set those headers. We should discuss the clumsiness here in general.

PR #74, opened 2017-11-14
Some CSS updates

mbrandonw:
Needed a few things, will make comments inline!

### DIFF
--- a/Sources/Css/Flexbox.swift
+++ b/Sources/Css/Flexbox.swift
@@ -145,7 +145,7 @@ public struct FlexWrap: Val, Other {
 }
 
 public func flex(
-  grow: Int,
+  grow: Int? = nil,
   shrink: Int? = nil,
   basis: Size? = nil
   )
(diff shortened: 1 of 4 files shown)


Sources/Css/Flexbox.swift
```diff
@@ -145,15 +145,15 @@ public struct FlexWrap: Val, Other {
 }
 
 public func flex(
-  grow: Int,
+  grow: Int? = nil,
```
mbrandonw: I thought this was required but it's not...

Sources/Css/Render.swift
```diff
@@ -33,7 +33,7 @@ func render(attribute: CssSelector.Attribute) -> String {
   }
 }
 
-func renderSelector(_ config: Config, _ selector: CssSelector) -> String {
+public func renderSelector(_ selector: CssSelector) -> String {
```
mbrandonw: I only wanted this public so that I could create a `class` helper for HTML attributes that took an array of `CssSelector`s, but also saw that `config` wasn't even being used.

PR #78, opened 2017-11-29
Move some todos over.

mbrandonw:
I moved the bare minimum of `TODO`s over from https://www.github.com/pointfreeco/pointfreeco so that it will build without `@testable imports`.

### DIFF
--- a/Sources/Css/Stylesheet.swift
+++ b/Sources/Css/Stylesheet.swift
@@ -130,15 +130,15 @@ extension Stylesheet: Monoid {
   }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/Css/Stylesheet.swift
```diff
@@ -130,15 +130,15 @@ extension Stylesheet: Monoid {
   }
 }
 
-func key<V: Val>(_ key: Key<V>, _ a: V) -> Stylesheet {
+public func key<V: Val>(_ key: Key<V>, _ a: V) -> Stylesheet {
```
mbrandonw: making these public so that we have escape hatches to css styles that we do not yet support.

PR #79, opened 2017-11-29
Fix crashes in URLComponents

mbrandonw:
We are seeing a lot of crashes stemming from this bug https://bugs.swift.org/browse/SR-6422, and they all seem to be related to `URLQueryItem`s. In this PR I have replaced a usage of it with our `urlFormEncode` helper and it fixed the crash :/


Dockerfile
```diff
@@ -4,6 +4,7 @@ WORKDIR /package
 
 COPY . ./
 
+RUN rm -rf /package/.build/debug
```
mbrandonw: this seems to have fixed the symlink error we get when run tests on linux+docker.

Package.swift
```diff
@@ -19,6 +19,7 @@ let package = Package(
     .library(name: "HttpPipelineHtmlSupport", targets: ["HttpPipelineHtmlSupport"]),
     .library(name: "HttpPipelineTestSupport", targets: ["HttpPipelineTestSupport"]),
     .library(name: "MediaType", targets: ["MediaType"]),
+    .library(name: "UrlFormEncoding", targets: ["UrlFormEncoding"]),
```
mbrandonw: I split out the `urlFormEncode` helper into its own package so that we could use it in a few spots...

Sources/ApplicativeRouter/SomePartialIsos.swift
```diff
@@ -143,7 +144,5 @@ private func formEncodedStringToFields(_ body: String) -> [String: String] {
 }
 
 private func fieldsToFormEncodedString(_ data: [String: String]) -> String {
-  let t = URLComponents()
-    |> \.queryItems .~ data.map(URLQueryItem.init(name:value:))
-  return t.query ?? ""
+  return urlFormEncode(value: data)
```
mbrandonw: this seems to be the cause of all of our crashes.

mbrandonw:
See also:

https://bugs.swift.org/browse/SR-6398
https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20171127/001376.html

Dockerfile
```diff
@@ -4,6 +4,9 @@ WORKDIR /package
 
 COPY . ./
 
+# Helps with: https://bugs.swift.org/browse/SR-6500
+RUN rm -rf /package/.build/debug
```
mbrandonw: this seems to have fixed the symlink error we get when run tests on linux+docker.

PR #82, opened 2017-12-03
Getting CommonCrypto and playgrounds to play nicely

mbrandonw:
### What

Right now playgrounds are busted cause of commoncrypto. Why is commoncrypto even in this repo? Good question! It probably shouldn't! But we'll move it out another day.

In the meantime, if you copy a module map to a particular location on your computer, it will all work.

Even tho commoncrpyto should probably be removed from this repo, this work will still be pertinent cause we need to do something similar in pointfreeco.


Makefile
```diff
@@ -13,6 +13,13 @@ imports = \
 xcodeproj:
 	swift package generate-xcodeproj
 
+bootstrap: common-crypto-mm xcodeproj
+
+common-crypto-mm:
+	-@mkdir "$(COMMON_CRYPTO_PATH)" || true
+	-@touch "$(COMMON_CRYPTO_MODULE_PATH)" || true
+	-@echo "$$COMMON_CRYPTO_MODULE_MAP" > "$(COMMON_CRYPTO_MODULE_PATH)"
```
mbrandonw: so this creates the `CommonCrypto.framework` directory, then adds a module map file, and then puts the module map code into the file.

Makefile
```diff
@@ -43,3 +50,16 @@ test-swift:
 	swift test
 
 test-all: test-linux test-mac test-ios
+
+SDK_PATH = $(shell xcrun --show-sdk-path)
```
mbrandonw: the place we install the module map is dependent on what Xcode you have selected (via `xcode-select`)

Web.playground/Pages/HttpPipeline.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+import Html
+import HttpPipeline
```
mbrandonw: `HttpPipeline` is the framework with the commoncrypto dependency, and so importing this used to break playgrounds. now it all works!

Makefile
```diff
@@ -13,6 +13,13 @@ imports = \
 xcodeproj:
 	swift package generate-xcodeproj
 
+bootstrap: common-crypto-mm xcodeproj
+
+common-crypto-mm:
+	-@mkdir "$(COMMON_CRYPTO_PATH)" || true
```
stephencelis: `mkdir -p` should negate the need for `|| true`

Makefile
```diff
@@ -13,6 +13,13 @@ imports = \
...
+	-@touch "$(COMMON_CRYPTO_MODULE_PATH)" || true
```
stephencelis: This should never fail, no? And if you're `echo`ing the next line, maybe it's not even necessary?

Makefile
```diff
@@ -43,3 +50,16 @@ test-swift:
 	swift test
 
 test-all: test-linux test-mac test-ios
+
+SDK_PATH = $(shell xcrun --show-sdk-path)
+FRAMEWORKS_PATH = $(SDK_PATH)/System/Library/Frameworks
+COMMON_CRYPTO_PATH = $(FRAMEWORKS_PATH)/CommonCrypto.framework
+COMMON_CRYPTO_MODULE_PATH = $(COMMON_CRYPTO_PATH)/module.map
+define COMMON_CRYPTO_MODULE_MAP
```
stephencelis: Whohohoa! Makefile expert! Never seen this before!

Makefile
mbrandonw: lol very true

Makefile
mbrandonw: ha... yahhhhh

Makefile
mbrandonw: 💅

PR #84, opened 2017-12-04
Add Some HTML4 Table Support

stephencelis:
I think we may need to add proper DOCTYPE support, too, if we go down this road.


mbrandonw:
Niiiiice! I’ll make a sample email in this and see if I need anything else. I can take care of the PR form here!

PR #86, opened 2017-12-04
Work around empty query components bug.

mbrandonw:
I just filed this bug https://bugs.swift.org/browse/SR-6405 and this PR works around it.


Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -159,12 +159,21 @@ private func request(from data: RequestData) -> URLRequest? {
...
+
+
+
+  // Due to this bug https://bugs.swift.org/browse/SR-6527, if `URLComponents` doesn't contain any path or
+  // query information, it will failure to create a `URL`.
+  return
+    (
+      data.path.isEmpty && data.query.isEmpty
+        ? (base ?? URL(string: "/"))
+        : urlComponents(from: data).url(relativeTo: base)
```
mbrandonw: this was coming back `nil` on linux, so we have to first check if path+query are empty, and if so just default to the base url.

Tests/ApplicativeRouterTests/TestRouter.swift
```diff
@@ -4,6 +4,7 @@ import Either
 import Prelude
 
 enum Routes {
+  case home
```
mbrandonw: we we're seeing this bug in the tests cause the `root` route technically when to the url `/home`. I've now create a separate `home` route so that we can test both and fully root URL and something with path components.

PR #91, opened 2017-12-12
Add URL Form Decoder for Decodable

stephencelis:
Our router has been very naively converting form data and query strings to `[String: String]`, which leads to a lot of extra manual parsing. While form data is typically thought of as a dictionary-like structure, this is a lossy format and key collisions make `[(String, String)]` a more accurate format. Because of this, different frameworks and services have adopted different key and value encodings to accommodate things like arrays and nested structure.

The most agnostic parser may work as follows:

``` swift
parse("xs=1&xs=2") == ["xs": ["1", "2"]] // true
```

Meanwhile, other parsers support bracketed and index-bracketed values:

``` swift
parse("xs[]=1&xs[]=2&roots[][value]=hello&roots[][value]=world")
  == ["xs": ["1", "2"], "roots": [["value": "hello"], ["value": "world"]]]
// or, with indices:
parse("xs[0]=1&xs[1]=2&roots[1][value]=world&roots[0][value]=hello")
  == ["xs": ["1", "2"], "roots": [["value": "hello"], ["value": "world"]]]
```

Still—and beyond the scope of this PR at the time of submitting—other parsers may favor array values by comma-separation!

``` swift
parse("xs=1,2") == ["xs": ["1", "2"]]
```

Even after all this, you're left with a tree of `[String: Any]` where the `Any` node eventually reaches `String`, so you inevitably have to parse further to get the data into its proper format.

This PR leverages `Decodable` to automatically convert form data and query strings into structures that we can work with immediately. It makes a best effort to convert each string value to its expected type.

- `"1"` becomes `1` when an `Int` (or any other number type) is expected.
- `"2.3"` becomes `2.3` as a `Float` or `Double` accordingly.
- `Date`- and `Data`-decoding strategies exist, to automatically hydrate the expected type.
- For nested parsing strategies, deep and complex structures can be decoded!

I'll add a quick self-review to highlight some tested examples.

### DIFF
--- a/Sources/UrlFormEncoding/UrlFormDecoder.swift
+++ b/Sources/UrlFormEncoding/UrlFormDecoder.swift
@@ -618,7 +618,7 @@ private func pairs(_ query: String) -> [(String, String)] {
     .split(separator: "&")
     .map { (pairString: Substring) -> (name: String, value: String) in
       let pairArray = pairString.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
-        .flatMap { $0.removingPercentEncoding }
+        .flatMap { String($0).removingPercentEncoding }
       return (pairArray[0], pairArray.count == 2 ? pairArray[1] : "")
     }
     .sorted { $0.name < $1.name }
--- a/Sources/UrlFormEncoding/UrlFormDecoder.swift
+++ b/Sources/UrlFormEncoding/UrlFormDecoder.swift
@@ -1,0 +1,629 @@
+import Foundation
+import Optics
+import Prelude
+
+public final class UrlFormDecoder: Decoder {
+  private(set) var containers: [Any] = []
+  private var container: Any {
+    return containers.last!
+  }
+  public private(set) var codingPath: [CodingKey] = []
+  public var dataDecodingStrategy: DataDecodingStrategy = .deferredToData
+  public var dateDecodingStrategy: DateDecodingStrategy = .deferredToDate
+  public var parsingStrategy: ParsingStrategy = .accumulatePairs
+  public let userInfo: [CodingUserInfoKey: Any] = [:]
+
+  public init() {
+  }
+
+  public func decode<T: Decodable>(_ type: T.Type, from data: Data) throws -> T {
+    let container = self.parsingStrategy.strategy(String(decoding: data, as: UTF8.self))
+    self.containers.append(container)
+    defer { self.containers.removeLast() }
+    return try T(from: self)
+  }
+
+  private func unbox(_ value: Any, as type: Data.Type) throws -> Data {
+    guard let string = singleton(value) else {
+      throw Error.decodingError("Expected string data, got \(value)", self.codingPath)
+    }
+
+    switch self.dataDecodingStrategy {
+    case .deferredToData:
+      return try Data(from: self)
+    case .base64:
+      guard let data = Data(base64Encoded: string) else {
+        throw Error.decodingError("Expected base64-encoded data, got \(string)", self.codingPath)
+      }
+      return data
+    case let .custom(strategy):
+      guard let data = strategy(string) else {
+        throw Error.decodingError("Failed strategy when decoding data from \(string)", self.codingPath)
+      }
+      return data
+    }
+  }
+
+  private func unbox(_ value: Any, as type: Date.Type) throws -> Date {
+    guard let string = singleton(value) else {
+      throw Error.decodingError("Expected string date, got \(value)", self.codingPath)
+    }
+
+    switch self.dateDecodingStrategy {
+    case .deferredToDate:
+      return try Date(from: self)
+    case .secondsSince1970:
+      guard let date = Double(string).map(Date.init(timeIntervalSince1970:)) else {
+        throw Error.decodingError("Expected seconds, got \(string)", self.codingPath)
+      }
+      return date
+    case .millisecondsSince1970:
+      guard let date = Double(string).map({ Date(timeIntervalSince1970: $0 / 1000) }) else {
+        throw Error.decodingError("Expected milliseconds, got \(string)", self.codingPath)
+      }
+      return date
+    case .iso8601:
+      let someDate = iso8601DateFormatter.date(from: string)
+        ?? iso8601DateFormatterWithoutMilliseconds.date(from: string)
+      guard let date = someDate else {
+        throw Error.decodingError("Expected ISO 8601 date string, got \(string)", self.codingPath)
+      }
+      return date
+    case let .formatted(formatter):
+      guard let date = formatter.date(from: string) else {
+        throw Error.decodingError("Expected \(formatter.dateFormat), got \(string)", self.codingPath)
+      }
+      return date
+    case let .custom(strategy):
+      guard let data = strategy(string) else {
+        throw Error.decodingError("Failed strategy when decoding data from \(string)", self.codingPath)
+      }
+      return data
+    }
+  }
+
+  private func unbox<T: Decodable>(_ value: Any, as type: T.Type) throws -> T {
+    if type == Data.self {
+      return try self.unbox(value, as: Data.self) as! T
+    } else if type == Date.self {
+      return try self.unbox(value, as: Date.self) as! T
+    } else {
+      return try T(from: self)
+    }
+  }
+
+  public func container<Key>(keyedBy type: Key.Type) throws
+    -> KeyedDecodingContainer<Key>
+    where Key: CodingKey {
+
+      guard let container = self.container as? [String: Any] else {
+        throw Error.decodingError("Expected keyed container, got \(self.container)", self.codingPath)
+      }
+      return .init(KeyedContainer(decoder: self, container: container))
+  }
+
+  public func unkeyedContainer() throws -> UnkeyedDecodingContainer {
+    guard let container = self.container as? [Any] else {
+      throw Error.decodingError("Expected unkeyed container, got \(self.container)", self.codingPath)
+    }
+    return UnkeyedContainer(decoder: self, container: container, codingPath: self.codingPath)
+  }
+
+  public func singleValueContainer() throws -> SingleValueDecodingContainer {
+    return SingleValueContainer(decoder: self, container: self.container)
+  }
+
+  public enum Error: Swift.Error {
+    case decodingError(String, [CodingKey])
+  }
+
+  struct KeyedContainer<Key: CodingKey>: KeyedDecodingContainerProtocol {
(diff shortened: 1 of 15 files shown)


Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,629 @@
...
+    if type == Data.self {
+      return try self.unbox(value, as: Data.self) as! T
+    } else if type == Date.self {
+      return try self.unbox(value, as: Date.self) as! T
+    } else {
+      return try T(from: self)
+    }
+  }
+
+  public func container<Key>(keyedBy type: Key.Type) throws
```
stephencelis: Quick NB for folks new to decoder development: there are 3 sub-types ("containers") that help a decoder decode:

- "Keyed" containers decode raw data into things that have stringly-typed "keys" (structs and their named properties, dictionaries with string keys).
- "Unkeyed" containers decode arrays of values.
- "Single value" containers decode...a single value.

In this world of form data, we first parse our string into `[String: Any]`, handling arrays and nested dictionaries with whatever strategy our decoder picks. Once in this realm, we traverse the structure, handing each node off to the appropriate container implicitly:

- `[String: Any]` sub-nodes get sent to a "keyed" container.
- `[Any]` sub-nodes get sent to an "unkeyed" container.
- `Any`thing else gets sent to a single-value container.

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,629 @@
...
+    }
+
+    func decode<T>(_ type: T.Type) throws -> T where T: Decodable {
+      self.decoder.containers.append(self.container)
+      defer { self.decoder.containers.removeLast() }
+      return try self.decoder.unbox(self.container, as: T.self)
+    }
+  }
+
+  public enum DataDecodingStrategy {
```
stephencelis: These decoding strategies likewise mirror `JSONDecoder`'s, with one difference: `JSONDecoder.custom` is `(Decoder) throws -> Data`, while we're `(String) -> Data?`. We lose some error messaging ability (which we could regain with `(String) throws -> Data`, but things are generally more flexible. We can always capture a custom decoder locally and pass it our string.

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,629 @@
...
+
+private let iso8601DateFormatter = DateFormatter()
+  |> iso8601
+  |> \.dateFormat .~ "yyyy-MM-dd'T'HH:mm:ss.SSSXXXXX"
+
+private let iso8601DateFormatterWithoutMilliseconds = DateFormatter()
+  |> iso8601
+  |> \.dateFormat .~ "yyyy-MM-dd'T'HH:mm:ssXXXXX"
+
+private func parse(isArray: @escaping (String) -> Bool) -> (String) -> [String: Any] {
```
stephencelis: From here to the end of this file is an implementation of nested query parameter parsing. It's gnarly and a little imperative (I hope there aren't too many bugs!). Ideas for cleanup welcome!

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,629 @@
...
+    .split(separator: "&")
+    .map { (pairString: Substring) -> (name: String, value: String) in
+      let pairArray = pairString.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+        .flatMap { String($0).removingPercentEncoding }
+      return (pairArray[0], pairArray.count == 2 ? pairArray[1] : "")
+    }
+    .sorted { $0.name < $1.name }
+}
+
+private func singleton(_ value: Any) -> String? {
```
stephencelis: Oh, one other thing in this file! It seems safest (and was easier to implement) to decode `A` or `[A]` depending on the data structure without worrying about the query strategy. Happy to discuss any concerns, though!

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
```diff
@@ -0,0 +1,133 @@
+import Prelude
+import SnapshotTesting
+import UrlFormEncoding
+import XCTest
+
+final class UrlFormDecoderTests: XCTestCase {
```
stephencelis: These tests provide a bunch of examples! Would be fun to get some kind of property testing here, though.

stephencelis:
Not sure what the Circle failures are about but going to take a breather. Any ideas? Wonder if we can redirect the raw output to a Circle artifact or disable xcpretty so we can debug a bit more easily.

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
mbrandonw: we gotta add these to the `imports` var in the `Makefile` for linux :/ just did it here 96dab4b

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,626 @@
...
+  }
+
+  public init?(intValue: Int) {
+    self.init(index: intValue)
+  }
+}
+
+private let iso8601 = ((\DateFormatter.calendar) .~ Calendar(identifier: .iso8601))
+  >>> ((\DateFormatter.locale) .~ Locale(identifier: "en_US_POSIX"))
+  >>> ((\DateFormatter.timeZone) .~ TimeZone(abbreviation: "UTC"))
```
mbrandonw: beaut

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,626 @@
...
+      var params: [String: Any] = [:]
+      for (name, value) in pairs(query) {
+        var values = params[name] as? [Any] ?? []
+        values.append(value)
+        params[name] = values
+      }
+      return params
+    }
+
+    public static let brackets = ParsingStrategy.custom(parse(isArray: ^\.isEmpty))
```
mbrandonw: we can even drop `ParsingStrategy.` yeah?

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -0,0 +1,626 @@
...
+
+private func pairs(_ query: String) -> [(String, String)] {
+  return query
+    .split(separator: "&")
+    .map { (pairString: Substring) -> (name: String, value: String) in
+      let pairArray = pairString.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+        .flatMap { String($0).removingPercentEncoding }
+      return (pairArray[0], pairArray.count == 2 ? pairArray[1] : "")
+    }
+    .sorted { $0.name < $1.name }
```
mbrandonw: if I remember correctly, `sorted` is not a stable sort, which means it's possible that the string `x[]=1&x[]=2` could get sorted into pairs `("x[]", "2"), ("x[]", "1")`. theoretically at least.

we can easily fix by doing `.enumerated().sorted().map(first)` or something. do you think it's worth it?

Sources/UrlFormEncoding/UrlFormDecoder.swift
mbrandonw: not sure I understand. could you explain a bit more?

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  func testDataDecodingWithBase64() throws {
+    struct MyData: Decodable {
+      let data: Data
+    }
+
+    decoder.dataDecodingStrategy = .base64
+    XCTAssertEqual(
+      "OOPs",
```
mbrandonw: lol

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
```diff
@@ -0,0 +1,133 @@
...
+    decoder.dateDecodingStrategy = .iso8601
+    assertSnapshot(
+      matching: try decoder.decode(MyDate.self, from: Data("date=2017-12-11T20:36:00.000-05:00".utf8))
+    )
+    assertSnapshot(
+      matching: try decoder.decode(MyDate.self, from: Data("date=2017-12-11T20:36:00-05:00".utf8))
+    )
+  }
+
+//  func testDateDecodingWithFormatted() throws {
```
mbrandonw: we keeping this?

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
mbrandonw: oops never mind, I thought this was in a new module! reverted...

Sources/UrlFormEncoding/UrlFormDecoder.swift
stephencelis: Ah, an example would help! Given the following struct:

``` swift
struct Box: Decodable {
  let a: [Int]
  let b: Int
}
```

And the following query string:

```
a=1&b=1
```

The underlying data representation is `["a": ["1"], "b", ["1"]]`, so `b` can unbox the value using this helper. It's mostly a concession for the default representation, though, so maybe its logic should move to the strategy somehow?

Sources/UrlFormEncoding/UrlFormDecoder.swift
stephencelis: Could we use our comparator for a stable sort? I'm surprised that sorting a list of people by their name may shuffle two "Margaret"s.

Tests/UrlFormEncodingTests/UrlFormDecoderTests.swift
stephencelis: Might come up with a better test. I think Linux/Mac just aren't consistent here.

Sources/UrlFormEncoding/UrlFormDecoder.swift
mbrandonw: > Could we use our comparator for a stable sort? 

Anything that uses `Array.sorted` under the hood (which our comparators do via helpers) is susceptible to unstable sorts. We can easily make a `stablySorted` that does the enumerated trick. And if we further have lenses on tuples, and comparators on lenses we could do something like (pseudo code):

```swift 
stablySorted = enumerated
  .sorted(second.comparator <> first.comparator)
  .map(second)
```

stephencelis:
Merging since it passes locally and on Travis! Will hopefully resolve the Circle issue when a newer version of Xcode is available.

PR #92, opened 2017-12-13
URL form decoder tweaks

stephencelis:
Just making the strategies we use a bit more honest and documenting them along the way. Wrote these while waiting for food and just committed em now. Will probably merge so I can pin `PointFree` to the latest on my branch!

### DIFF
--- a/Sources/UrlFormEncoding/UrlFormDecoder.swift
+++ b/Sources/UrlFormEncoding/UrlFormDecoder.swift
@@ -17,14 +17,30 @@ public final class UrlFormDecoder: Decoder {
   }
 
   public func decode<T: Decodable>(_ type: T.Type, from data: Data) throws -> T {
-    let container = self.parsingStrategy.strategy(String(decoding: data, as: UTF8.self))
+    let query = String(decoding: data, as: UTF8.self)
+    let container: [String: Any]
+    switch self.parsingStrategy {
+    case .accumulatePairs:
+      container = accumulatePairs(query)
+    case let .custom(strategy):
+      container = strategy(query)
+    }
     self.containers.append(container)
     defer { self.containers.removeLast() }
     return try T(from: self)
   }
 
+  private func unbox(_ value: Any) -> String? {
+    switch self.parsingStrategy {
+    case .accumulatePairs:
+      return value as? String ?? (value as? [String])?.last
+    case .custom:
+      return value as? String
+    }
+  }
+
   private func unbox(_ value: Any, as type: Data.Type) throws -> Data {
-    guard let string = singleton(value) else {
+    guard let string = unbox(value) else {
       throw Error.decodingError("Expected string data, got \(value)", self.codingPath)
     }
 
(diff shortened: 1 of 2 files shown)


Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -10,21 +10,37 @@ public final class UrlFormDecoder: Decoder {
   public private(set) var codingPath: [CodingKey] = []
   public var dataDecodingStrategy: DataDecodingStrategy = .deferredToData
   public var dateDecodingStrategy: DateDecodingStrategy = .deferredToDate
-  public var parsingStrategy: ParsingStrategy = .accumulatePairs
+  public var parsingStrategy: ParsingStrategy = .accumulateValues
```
stephencelis: Don't really know why I named it the other way. Was dealing with a lot of pairs at the time.

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -10,21 +10,37 @@ public final class UrlFormDecoder: Decoder {
...
 
   public init() {
   }
 
   public func decode<T: Decodable>(_ type: T.Type, from data: Data) throws -> T {
-    let container = self.parsingStrategy.strategy(String(decoding: data, as: UTF8.self))
+    let query = String(decoding: data, as: UTF8.self)
+    let container: [String: Any]
+    switch self.parsingStrategy {
+    case .accumulateValues:
```
stephencelis: Kinda funny to get here after that struct vs. enum convo last week! The one interesting thing that ya lose when moving from an `enum` to a `struct` that wraps a closure is the ability to tag and switch on the cases!

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -10,21 +10,37 @@ public final class UrlFormDecoder: Decoder {
...
+    }
     self.containers.append(container)
     defer { self.containers.removeLast() }
     return try T(from: self)
   }
 
+  private func unbox(_ value: Any) -> String? {
+    switch self.parsingStrategy {
+    case .accumulateValues:
+      return value as? String ?? (value as? [String])?.last
```
stephencelis: We now only auto-unbox array values for the default, flat way.

stephencelis:
Wow, looks like Linux and macOS have different behavior for `Array.removeLast`! Gonna file a bug.

stephencelis:
On closer inspection, the behavior's the same! Not sure what is preventing the crash on macOS, but filed a bug here: https://bugs.swift.org/browse/SR-6598

Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -510,24 +526,71 @@ public final class UrlFormDecoder: Decoder {
...
-      for (name, value) in pairs(query) {
-        var values = params[name] as? [Any] ?? []
-        values.append(value)
-        params[name] = values
-      }
-      return params
-    }
-
+  public enum ParsingStrategy {
+    /// A parsing strategy that accumulates values when multiple keys are provided.
```
mbrandonw: amazing docs! we need to do this more after launch and use jazzy or something to generate docs

PR #93, opened 2017-12-14
More query things

stephencelis:
This ties our new decoder (#91) to our router. There are a few TODO items for things like pretty printing, but this makes things usable!

### DIFF
--- a/Sources/ApplicativeRouter/Combinators.swift
+++ b/Sources/ApplicativeRouter/Combinators.swift
@@ -13,8 +13,8 @@ public func lit(_ str: String) -> Router<Prelude.Unit> {
             : nil
       }
   },
-    print: { _ in .init(method: nil, path: [str], query: [:], body: nil) },
-    template: { _ in .init(method: nil, path: [str], query: [:], body: nil) }
+    print: { _ in .init(method: nil, path: [str], query: nil, body: nil) },
+    template: { _ in .init(method: nil, path: [str], query: nil, body: nil) }
   )
 }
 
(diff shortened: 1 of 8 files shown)


Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -13,8 +14,8 @@ public func lit(_ str: String) -> Router<Prelude.Unit> {
             : nil
       }
   },
-    print: { _ in .init(method: nil, path: [str], query: [:], body: nil) },
-    template: { _ in .init(method: nil, path: [str], query: [:], body: nil) }
+    print: { _ in .init(method: nil, path: [str], query: nil, body: nil) },
```
stephencelis: First thing you may notice is that `query` is no longer `[String: String]`. We do less parsing here (so that `UrlFormEncoding` can do it for us) and leave `query` as `String?`.

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -44,20 +45,27 @@ public func pathParam<A>(_ f: PartialIso<String, A>) -> Router<A> {
 public func queryParam<A>(_ key: String, _ f: PartialIso<String?, A>) -> Router<A> {
   return .init(
     parse: { route in
-      return f.apply(route.query[key]).map { (route, $0) }
+      return f.apply(route.params[key]).map { (route, $0) }
```
stephencelis: You'll see, later, that `params` is a computed property converting the `String?` query into `[String: String]`. We can definitely optimize here when need be.

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -44,20 +45,27 @@ public func pathParam<A>(_ f: PartialIso<String, A>) -> Router<A> {
...
+      let params = f.unapply(a).flatMap { $0.map { [key: $0] } }
+      return RequestData(method: nil, path: [], query: params.map(print(params:)), body: nil)
     },
     template: { a in
-      RequestData(method: nil, path: [], query: [key: ":\(typeKey(a))"], body: nil)
+      RequestData(method: nil, path: [], query: print(params: [key: ":\(typeKey(a))"]), body: nil)
   })
 }
 
+private func print(params: [String: String]) -> String {
```
stephencelis: Not sure if there's a better place for this.

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -125,30 +133,28 @@ extension Router {
 }
 
 /// Parses the query params to create a value of type `A` via the `Codable` protocol.
-/// TODO: this only works for types `A` with string fields. Is it possible to improve that?
```
stephencelis: `queryParams` is now fully codable! We can give it a type that has non-string fields!

Sources/ApplicativeRouter/Combinators.swift
```diff
@@ -174,7 +180,7 @@ public let put = method(.put)
 
 // MARK: - Private
 
-private func typeKey<A>(_ a: A) -> String {
+private func typeKey(_ a: Any) -> String {
```
stephencelis: Without this, the printer prints `:any` for `queryParams` values. With it, it prints `:__nscfnumber` for int values...progress?

Tests/ApplicativeRouterTests/TestRouter.swift
```diff
@@ -34,11 +39,16 @@ let router: Router<Routes> = [
...
-  // /path/to/somewhere/cool?ref=:optional_string&active=:bool&t=:int
+  // GET /path/to/somewhere/cool?ref=:optional_string&active=:bool&t=:int
   Routes.iso.simpleQueryParams
     <¢> get %> lit("path") %> lit("to") %> lit("somewhere") %> lit("cool")
     %> queryParam("ref", opt(.string)) <%> queryParam("active", .bool) <%> queryParam("t", .int)
     <% end,
+
+  // GET /subscribe?plan=:int&token=:int
+  Routes.iso.codableQueryParams
+    <¢> get %> lit("subscribe") %> queryParams(SubscribeData.self)
```
stephencelis: This type could be inferred, though we also have a custom decoder to worry about, so the inconsistency between a computed `queryParam` vs. default decoder `queryParam()` vs. `queryParam(myDecoder)` is a pain.

stephencelis:
Local linux tests are failing with our old friend :(

```
Test Suite 'SyntaxRouterTests' started
Test Case 'SyntaxRouterTests.testHome' started
Fatal error: Constant strings cannot be deallocated: file Foundation/NSCFString.swift, line 118
```

Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -171,7 +171,8 @@ private func request(from data: RequestData, base: URL?) -> URLRequest? {
 }
 
 private func urlComponents(from route: RequestData) -> URLComponents {
-  return URLComponents()
-    |> \.path .~ route.path.joined(separator: "/")
```
stephencelis: @mbrandonw this lensing crashes Linux :/ May be fixed on the next version, but wonder if we should file a bug?

Tests/ApplicativeRouterTests/SyntaxRouterTests.swift
```diff
@@ -148,9 +148,16 @@ class SyntaxRouterTests: XCTestCase {
...
-      router.templateUrl(for: route)?.absoluteString
-    )
+    #if os(Linux)
+      XCTAssertEqual(
+        "subscribe?plan=:int",
+        router.templateUrl(for: route)?.absoluteString
+      )
+    #else
+      XCTAssertEqual(
+        "subscribe?plan=:__nscfnumber", // FIXME
```
stephencelis: Fun how this works in Linux but not macOS. Not sure if there's a way to coerce these `NSValue`s into their primitives. We may need to add a gnarly hack for pretty printing.

PR #95, opened 2017-12-14
Don't render boolean attributes that are false.

mbrandonw:
Don't know if I solved this the right way, but it at least fixed my test!

### DIFF
--- a/Sources/Html/Html.swift
+++ b/Sources/Html/Html.swift
@@ -85,7 +85,7 @@ public func comment<T>(_ content: StaticString) -> ChildOf<T> {
 
 extension Value {
   public func render(with key: String) -> EncodedString? {
-    return self.renderedValue().map { Html.encode("\(key)=") + quote($0) } ?? Html.encode(key)
+    return self.renderedValue().map { Html.encode(key + "=") + quote($0) } ?? Html.encode(key)
   }
 }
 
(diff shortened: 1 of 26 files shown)


Tests/HtmlTests/HtmlTests.swift
```diff
@@ -206,4 +206,12 @@ class HTMLTests: XCTestCase {
 
     assertSnapshot(matching: doc)
   }
+
+  func testBooleanAttributes() {
+    let uncheckedInput = input([id("not-checked"), checked(false)])
+    XCTAssertEqual("<input id=\"not-checked\">", render(uncheckedInput))
```
mbrandonw: this test fails without the change

Sources/Html/Html.swift
```diff
@@ -85,6 +85,13 @@ public func comment<T>(_ content: StaticString) -> ChildOf<T> {
 
 extension Value {
   public func render(with key: String) -> EncodedString? {
+
+    if let boolValue = self as? Bool {
```
stephencelis: Should this be in `extension Bool: Value`?

Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -5,6 +5,12 @@ import Prelude
 import SnapshotTesting
 import XCTest
 
+extension Application {
+  var isOther: Bool {
+    return !["javascript", "json", "xml", "x-www-form-url-encoded"].contains(self.rawValue)
```
stephencelis: This is kinda a hack just for this file. Wonder if we can tackle the logic that depends on it in a different way.

stephencelis:
Merging this because it's late and I tiredly spent 30 min trying to figure out why the pricing page wasn't changing with boolean values. Duh.

PR #96, opened 2017-12-16
More Form Data Codable Things

stephencelis:
A few things that have come up on my Point-Free branch.

PR #97, opened 2017-12-16
Optional codable query params should route

stephencelis:
Fixed decoding in #96, but still need this fix for routing.


stephencelis:
Good thing we have tests! Don't seem to have this quite right yet. Grabbin food and figuring out in a bit.

PR #103, opened 2017-12-28
HttpPipeline Response Cleanup

stephencelis:
Refamiliarizing myself with this stuff over coffee and decided to adjust some things.


Sources/HttpPipeline/Response.swift
```diff
@@ -1,7 +1,146 @@
...
+import MediaType
+import Prelude
 
 public struct Response {
   public private(set) var status: Status
-  public private(set) var headers: [ResponseHeader]
+  public private(set) var headers: [Header]
   public private(set) var body: Data
+
+  public struct Header {
```
stephencelis: Nesting this now (instead of hiding away in another file).

`ResponseHeader` used to be an `enum`. Now it's a `struct`. This means that we eagerly convert values to strings. (Is this potentially extra work for a failing request? Is this worth making lazy?) This also means we can't pattern-match anymore, which is fine, since `case let .contentType(t), let .other("Content-Type", t)` doesn't really work.

Sources/HttpPipeline/Response.swift
```diff
@@ -1,7 +1,146 @@
...
+
+    public static func contentType(_ type: MediaType) -> Header {
+      return .init("Content-Type", type.description)
+    }
+
+    public static func location(_ location: String) -> Header {
+      return .init("Location", location)
+    }
+
+    public static func setCookie(_ name: String, _ value: String, _ options: Set<CookieOption> = [])
```
stephencelis: I've removed the need for parameter names here. Autocompletion will give `<# name #>` and `<# value #>` hints in this post-`enum` world, and order seems to retain enough semantics here.

Tests/HttpPipelineTests/__Snapshots__/HttpPipelineTests/testCookieOptions.1.Conn.txt
```diff
@@ -17,7 +17,7 @@
   Set-Cookie: foo=bar; Max-Age=3600
   Set-Cookie: foo=bar; Path=/path/to/some/where
   Set-Cookie: foo=bar; SameSite=Lax
-  Set-Cookie: foo=bar; SameSite=String
```
stephencelis: Oops.

Tests/HttpPipelineTests/HttpPipelineTests.swift
```diff
@@ -63,16 +63,16 @@ class HttpPipelineTests: XCTestCase {
...
       writeStatus(.ok)
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.domain("www.pointfree.co")]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.expires(1234567890)]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.httpOnly]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.maxAge(3600)]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.path("/path/to/some/where")]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.sameSite(.lax)]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.sameSite(.strict)]))
-        >-> writeHeader(.setCookie(key: "foo", value: "bar", options: [.secure]))
+        >-> writeHeader(.setCookie("foo", "bar", [.domain("www.pointfree.co")]))
```
stephencelis: I think it reads pretty nice!

Tests/HttpPipelineTests/HttpPipelineTests.swift
```diff
@@ -52,9 +52,9 @@ class HttpPipelineTests: XCTestCase {
   func testCookies() {
     let middleware: Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data> =
       writeStatus(.ok)
-        >-> writeHeader(.setCookie(key: "user_id", value: "123456", options: []))
-        >-> writeHeader(.setCookie(key: "lang", value: "es", options: []))
-        >-> writeHeader(.clearCookie(key: "test"))
+        >-> writeHeader(.setCookie("user_id", "123456"))
```
stephencelis: Nice to not have to specify options!

Tests/HttpPipelineTests/__Snapshots__/HttpPipelineTests/testCookieOptions.1.Conn.txt
```diff
@@ -17,7 +17,7 @@
...
+  Set-Cookie: foo=bar; SameSite=Strict
```
mbrandonw: ha oops!

Sources/HttpPipeline/Response.swift
```diff
@@ -1,7 +1,146 @@
...
+      case let .basic(realm: realm):
+        return .init("WWW-Authenticate", "Basic" + (realm.map { " realm=\"" + $0 + "\"" } ?? ""))
+      }
+    }
+
+    public enum Authenticate {
+      case basic(realm: String?)
+    }
+
+    public enum CookieOption: Hashable, CustomStringConvertible, Comparable {
```
mbrandonw: do you think it makes sense to convert all the enums to structs?

stephencelis: For `Authenticate` maybe, but it's simple enough now while we only support basic auth. For `CookieOption` it's a finite set so `enum` seems to work nicely here. Other thoughts?

Tests/HttpPipelineTests/HttpPipelineTests.swift
mbrandonw: yeah I like!

Sources/HttpPipeline/Response.swift
mbrandonw: yeah makes sense. just re-questioning _all_ the enums these days

Sources/HttpPipeline/Response.swift
```diff
@@ -129,7 +129,7 @@ public struct Response {
     }
 
     public static func clearCookie(_ name: String) -> Header {
-      return .setCookie(name, "", [CookieOption.maxAge(0), .expires(0)])
```
stephencelis: Since we were operating on `TimeInterval` here, Swift seemed confused without the qualifying `CookieOption`. Changing `expires` to take a `Date` prevents this ambiguity (which is prob better for the API). We also immediately convert to a `Date` anyway.

Sources/HttpPipeline/Response.swift
stephencelis: Same! Lol.

PR #105, opened 2017-12-29
Some expression time fixes

stephencelis:
I think the biggest offenders are in `PointFree`, but these were getting flagged while I was in there.

### DIFF
--- a/Sources/Css/Color.swift
+++ b/Sources/Css/Color.swift
@@ -7,8 +7,8 @@ public enum Color: Val, Auto, Inherit, None {
   public func value() -> Value {
     switch self {
     case let .rgba(red, green, blue, alpha):
-      let formatted = alpha == 1.0
-        ? "#" + phex(red) + phex(green) + phex(blue)
+      let formatted: String = alpha == 1.0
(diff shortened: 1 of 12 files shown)


Sources/Css/Color.swift
```diff
@@ -7,8 +7,8 @@ public enum Color: Val, Auto, Inherit, None {
   public func value() -> Value {
     switch self {
     case let .rgba(red, green, blue, alpha):
-      let formatted = alpha == 1.0
-        ? "#" + phex(red) + phex(green) + phex(blue)
```
stephencelis: These operators had a cost :(

Sources/MediaType/MediaType.swift
```diff
@@ -45,9 +45,11 @@ public enum MediaType: CustomStringConvertible {
     case let .model(model):
       return "model/" + model
     case let .multipart(multipart, boundary):
-      return "multipart/" + multipart.rawValue + (boundary.map { "; boundary=" + $0 } ?? "")
```
stephencelis: Seems like this common interplay of `+` and `??` take awhile to typecheck.

PR #106, opened 2017-12-30
Some todos and fixes

mbrandonw:
We have quite a few todos we need to move over from pointfreeco, but these are what we've needed most recently, and this contains a CSS bug fix.

### DIFF
--- a/Development.xcconfig
+++ b/Development.xcconfig
@@ -1,1 +1,1 @@
-OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=80
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=400
--- a/Sources/Css/Color.swift
+++ b/Sources/Css/Color.swift
@@ -16,10 +16,8 @@ public enum Color: Val, Inherit, Initial {
     }
   }
 
-  public static let auto: Color = .other(.auto)
   public static let inherit: Color = .other(.inherit)
   public static let initial: Color = .other(.initial)
-  public static let none: Color = .other(.none)
   public static let transparent: Color = .other("transparent")
 
   public static func rgb(_ red: UInt8, _ green: UInt8, _ blue: UInt8) -> Color {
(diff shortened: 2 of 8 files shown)


Development.xcconfig
```diff
@@ -1 +1 @@
-OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=80
+OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=400
```
mbrandonw: bumping to what we use in pointfreeco

Sources/Css/Color.swift
```diff
@@ -1,6 +1,6 @@
 import Prelude
 
-public enum Color: Val, Auto, Inherit, None {
+public enum Color: Val, Inherit, Initial {
```
mbrandonw: turns out `auto` and `none` are not valid colors

Sources/Css/Display.swift
```diff
@@ -214,8 +214,11 @@ public func rect(top: Css.Size, right: Css.Size, bottom: Css.Size, left: Css.Siz
       [
         "rect(",
         top.value().unValue,
+        ",",
```
mbrandonw: forgot to intersperse commas between clip values. it's now rendering as `rect(1px1px1px1px)` causing a css error

Sources/HttpPipeline/SignedCookies.swift
```diff
@@ -105,18 +105,14 @@ extension Response.Header {
   }
 }
 
-private func digest(value: String, secret: String) -> String? {
+public func digest(value: String, secret: String) -> String? {
```
mbrandonw: we've needed some of these encryption/digest methods to be public

Development.xcconfig
stephencelis: I think everything here is 100ms or under. Would love to not let 400ms stuff sneak in if we can avoid it.

mbrandonw: I've got about 20 warnings that are over 100ms :/ it's prob all so dependent on the current state of your computer, but right now pretty much everything is definitely less than 400ms

PR #110, opened 2018-01-06
Plain Text Improvements

stephencelis:
There's a lot more we can do here, but this is a start!


stephencelis:
This is kinda an html2markdown solution.

PR #111, opened 2018-01-06
More plain-text fixes

stephencelis:
Might add a few more.


mbrandonw:
oh did you wanna go ahead an unescape `&nbsp;` while yer here?

PR #112, opened 2018-01-29
Fix bug in printing root string

mbrandonw:


Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -41,7 +41,8 @@ public struct Router<A> {
   }
 
   public func absoluteString(for a: A) -> String? {
-    return url(for: a).map { "/" + $0.absoluteString }
+    return (self.url(for: a)?.absoluteString)
+      .map { $0 == "/" ? "/" : "/" + $0 }
```
mbrandonw: a bit hacky, but this prevents the double `//` printing bug. think this is related to the `? (base ?? URL(string: "/"))` line in the `request(from: RequestData)` helper method.

this may be the right way to do it, or we can fix later...

Tests/ApplicativeRouterTests/SyntaxRouterTests.swift
```diff
@@ -28,6 +28,7 @@ class SyntaxRouterTests: XCTestCase {
     XCTAssertEqual(route, router.match(request: request))
     XCTAssertEqual(request, router.request(for: route, base: baseUrl))
     XCTAssertEqual("/", router.templateUrl(for: route)?.absoluteString)
+    XCTAssertEqual("/", router.absoluteString(for: .root))
```
mbrandonw: this test failed until the fix

PR #113, opened 2018-02-24
Bring in more TODOs

stephencelis:
Does what it says in the PR title.

### DIFF
--- a/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
+++ b/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
@@ -2,6 +2,34 @@ import Foundation
 import Optics
 import Prelude
 
+public func filterMap<A, B>(
+  _ f: @escaping (A) -> IO<B?>,
+  or notFoundMiddleware: @escaping Middleware<StatusLineOpen, ResponseEnded, A, Data>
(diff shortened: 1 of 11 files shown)


Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -46,30 +74,15 @@ public func notFound<A>(_ middleware: @escaping Middleware<HeadersOpen, Response
       >-> middleware
 }
 
-public func contentLength<A, B>(
```
stephencelis: Getting rid of this transformer since it doesn't seem to be useful...

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -2,6 +2,34 @@ import Foundation
 import Optics
 import Prelude
 
+public func filterMap<A, B>(
+  _ f: @escaping (A) -> IO<B?>,
```
stephencelis: Some food for thought, but I wonder if we should have a `filterMap`-like operation that goes `(Conn<A>) -> IO<B?>`... That way we could recreate some of the transformers below with it.

PR #114, opened 2018-03-03
Add rel and as to script tags.

mbrandonw:
just some new attributes for script tags.


mbrandonw:
oh oops, `link`! I got mixed up cause what I want to use this for is prism.js:

https://github.com/pointfreeco/pointfreeco/blob/6d20f53c082678439aadd975a9ad3e733f652c07/Sources/PointFree/PageLayout.swift#L192

mbrandonw:
this is the documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

Sources/Html/Attributes.swift
```diff
@@ -586,6 +588,18 @@ public func rel<T: HasRel>(_ value: Rel) -> Attribute<T> {
...
 
+public enum As: String {
+  case image
+  case font
+  case script
+  case style
+  case video
+}
+public protocol HasAs {}
+public func `as`<T: HasAs>(_ value: As) -> Attribute<T> {
```
stephencelis: If only one thing has `as` we can avoid the proto:

``` swift
public func `as`(_ value: As) -> Attribute<Element.Link> {
```

stephencelis:
Gonna close this out since I think they were addressed in swift-html :)

PR #119, opened 2018-03-30
4.1 release

stephencelis:
Locking down to the release.


Sources/Html/Render.swift
```diff
@@ -34,7 +34,7 @@ public func render(_ node: Node, config: Config = compact) -> String {
     case let .element(element):
       let _attribs = element.attribs
         .map { attrib in attrib.value.render(with: attrib.key)?.string }
-        .flatMap { $0 }
+        .compactMap { $0 }
```
stephencelis: No flatMap left behind.

PR #123, opened 2018-04-02
NIO Support

stephencelis:
This is a proof-of-concept and has some stuff to figure out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,15 @@ xcodeproj:
 	swift package generate-xcodeproj --xcconfig-overrides Development.xcconfig
 	xed .
 
-bootstrap: common-crypto-mm xcodeproj
+bootstrap: common-crypto-mm xcodeproj xcodeproj-mm
 
 common-crypto-mm:
 	-@sudo mkdir -p "$(COMMON_CRYPTO_PATH)"
 	-@echo "$$COMMON_CRYPTO_MODULE_MAP" | sudo tee "$(COMMON_CRYPTO_MODULE_MAP_PATH)" > /dev/null
+
+xcodeproj-mm:
+	-@ls Web.xcodeproj/GeneratedModuleMap | xargs -n1 -I '{}' sudo mkdir -p "$(FRAMEWORKS_PATH)/{}.framework"
+	-@ls Web.xcodeproj/GeneratedModuleMap | xargs -n1 -I '{}' sudo cp "./Web.xcodeproj/GeneratedModuleMap/{}/module.modulemap" "$(FRAMEWORKS_PATH)/{}.framework/module.map"
 
 linux-main:
 	sourcery \
--- a/Package.swift
+++ b/Package.swift
@@ -24,6 +24,7 @@ let package = Package(
   dependencies: [
     .package(url: "https://github.com/pointfreeco/swift-prelude.git", .revision("777966e")),
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing.git", .revision("c510e7d")),
+    .package(url: "https://github.com/apple/swift-nio.git", from: "1.0.0"),
     .package(url: "https://github.com/bkase/DoctorPretty.git", .exact("0.4.1")),
     .package(url: "https://github.com/IBM-Swift/BlueCryptor.git", .exact("0.8.22"))
   ],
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -14,12 +14,16 @@ xcodeproj:
...
 	xed .
 
-bootstrap: common-crypto-mm xcodeproj
+bootstrap: common-crypto-mm xcodeproj xcodeproj-mm
 
 common-crypto-mm:
 	-@sudo mkdir -p "$(COMMON_CRYPTO_PATH)"
 	-@echo "$$COMMON_CRYPTO_MODULE_MAP" | sudo tee "$(COMMON_CRYPTO_MODULE_MAP_PATH)" > /dev/null
 
+xcodeproj-mm:
```
stephencelis: This is a fun one. It installs any module maps that are generated with the Xcode project.

Sources/HttpPipeline/NIO.swift
```diff
@@ -0,0 +1,135 @@
...
+    let reqPart = self.unwrapInboundIn(data)
+
+    switch reqPart {
+    case let .head(header):
+      self.request = URL(string: header.uri).map { url -> URLRequest in
+        var req = URLRequest(url: url)
+          |> \.httpMethod .~ method(from: header.method)
+          |> \.allHTTPHeaderFields .~ header.headers.reduce(into: [:]) { $0[$1.name] = $1.value }
+        req.url = req.value(forHTTPHeaderField: "Host")
+          .flatMap { URL(string: "http://" + $0 + url.absoluteString) }
```
stephencelis: This logic is almost certainly wrong. Need to figure out how to properly construct the full request URL.

Sources/HttpPipeline/NIO.swift
```diff
@@ -0,0 +1,135 @@
...
+      _ = ctx.channel.write(HTTPServerResponsePart.body(.byteBuffer(buffer)))
+
+      _ = ctx.channel.writeAndFlush(HTTPServerResponsePart.end(nil)).then {
+        ctx.channel.close()
+      }
+    }
+  }
+}
+
+private func method(from method: HTTPMethod) -> String {
```
stephencelis: Maybe I'm missing something, but there didn't seem to be a way to get the string representation of this thing.

Sources/HttpPipeline/NIO.swift
```diff
@@ -0,0 +1,137 @@
...
+
+  do {
+    let reuseAddrOpt = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR)
+    let group = MultiThreadedEventLoopGroup(numThreads: System.coreCount)
+    let bootstrap = ServerBootstrap(group: group)
+      .serverChannelOption(ChannelOptions.backlog, value: 256)
+      .serverChannelOption(reuseAddrOpt, value: 1)
+      .childChannelInitializer { channel in
+        channel.pipeline.configureHTTPServerPipeline()
+          .then { channel.pipeline.add(handler: HTTPResponseCompressor()) }
```
stephencelis: This enables gzipped responses. Not sure if it's worth making this configurable or not or if it should be disabled in dev. Seems fine?

Makefile
mbrandonw: oh nice, can that replace our other modulemap copying?

Package.swift
```diff
@@ -16,14 +16,16 @@ let package = Package(
     .library(name: "HtmlTestSupport", targets: ["HtmlTestSupport"]),
     .library(name: "HtmlPrettyPrint", targets: ["HtmlPrettyPrint"]),
     .library(name: "HttpPipeline", targets: ["HttpPipeline"]),
+//    .executable(name: "HttpPipelineExample", targets: ["HttpPipelineExample"]),
```
mbrandonw: do we wanna leave these in?

stephencelis: I'd like to get it workin but didn't want it to block this PR.

Sources/HttpPipeline/NIO.swift
```diff
@@ -0,0 +1,143 @@
...
+  case .PROPPATCH: return "PROPPATCH"
+  case .SUBSCRIBE: return "SUBSCRIBE"
+  case .MKCALENDAR: return "MKCALENDAR"
+  case .MKACTIVITY: return "MKACTIVITY"
+  case .UNSUBSCRIBE: return "UNSUBSCRIBE"
+  case let .RAW(value): return value
+  }
+}
+
+private func headers(from headers: HTTPHeaders) -> [String: String] {
```
mbrandonw: this doesn't look used anymore

stephencelis: Whoops, yeah. Deleted on the branch I had goin before 4.1 but forgot to update this one.

PR #129, opened 2018-04-04
New build system and updated package resolved.

mbrandonw:
The `Package.resolved` change when I ran `make xcodeproj` so just committing it. Also started using the new build system.


stephencelis:
These changes snuck in with the NIO stuff.

PR #143, opened 2018-06-29
Xcode 10

stephencelis:
Adds support (with a few workarounds) for Xcode 10.


Development.xcconfig
```diff
@@ -1 +1 @@
-OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=400
+OTHER_SWIFT_FLAGS=$(inherited) -DSWIFT_PACKAGE -Xfrontend -warn-long-expression-type-checking=400
```
stephencelis: This is a bug that's been resolved upstream. We can probably revert this line in the next beta.

https://bugs.swift.org/browse/SR-7912

PR #145, opened 2018-07-01
Update Snapshot Testing/Support

stephencelis:
Bumps snapshot testing and cleans up a few things.


Sources/Css/Config.swift
```diff
@@ -7,31 +7,37 @@ public struct Config {
...
 
-public let pretty = Config(align: true,
-                           finalSemicolon: true,
-                           indentation: "  ",
-                           lbrace: "{",
-                           newline: "\n",
-                           rbrace: "}",
-                           sep: " ",
-                           warn: true)
+  public static let pretty = Config(
```
stephencelis: Moving these to statics to be a lil more consistent.

Sources/HtmlCssSupport/Support.swift
```diff
@@ -13,20 +13,25 @@ public func style<T>(_ style: Stylesheet) -> Attribute<T> {
   return .init("style", style)
 }
 
-public func style(_ attribs: [Attribute<Element.Style>], _ css: Stylesheet) -> ChildOf<Element.Head> {
-  return style(render(config: Css.compact, css: css))
+public func style(
+  _ attribs: [Attribute<Element.Style>],
+  _ css: Stylesheet,
+  config: Css.Config = .compact
```
stephencelis: Exposing `config` publicly so we can render these in a pretty way for snapshots.

Sources/HttpPipeline/Status.swift
```diff
@@ -1,69 +1,140 @@
-public enum Status: Int {
-  case `continue` = 100
-  case switchingProtocols = 101
-  case processing = 102
+public struct Status: RawRepresentable {
```
stephencelis: Finally makin this a struct :)

Sources/HttpPipelineTestSupport/HttpPipelineTestSupport.swift
```diff
@@ -38,17 +38,11 @@ extension Response: Snapshot {
 
 extension Conn: Snapshot {
   public var snapshotFormat: String {
```
stephencelis: Cleaning up the snapshot format a bit to make it look more like a raw request/response.

PR #146, opened 2018-07-02
Fix pretty renderer for "pre" elements

stephencelis:
This fixes pretty HTML rendering for `pre` elements, which need to not be prefixed with whitespace since they preserve it.


stephencelis:
Yeeeah. I'll add tests before mergin. We prob definitely need to update the Doctor Pretty version. I'm less familiar with it but need to figure it out.

PR #155, opened 2018-10-03
Query should be an array, not dictionary

stephencelis:


Sources/ApplicativeRouter/SyntaxRouter.swift
```diff
@@ -204,7 +203,7 @@ private func urlComponents(from route: RequestData) -> URLComponents {
 
   if !route.query.isEmpty {
     components.queryItems = route.query
-      .sorted { lhs, rhs in lhs.key < rhs.key }
+      .filter { $0.value != nil }
```
stephencelis: Without this filtering, we get the key name for matches during printing.

Tests/ApplicativeRouterTests/SyntaxRouterTests.swift
```diff
@@ -119,15 +119,15 @@ class SyntaxRouterTests: XCTestCase {
...
+    let request = URLRequest(url: URL(string: "path/to/somewhere/cool?ref=hello&active=true&t=122")!)
       // NB: necessary for linux tests: https://bugs.swift.org/browse/SR-6405
       |> \.httpMethod .~ "get"
     let route = Routes.simpleQueryParams(ref: "hello", active: true, t: 122)
 
     XCTAssertEqual(route, router.match(request: request))
     XCTAssertEqual(request, router.request(for: route))
     XCTAssertEqual(
-      "path/to/somewhere/cool?active=:bool&ref=:optional_string&t=:int",
+      "path/to/somewhere/cool?ref=:optional_string&active=:bool&t=:int",
```
stephencelis: Order's preserved now! Which is what we want.

PR #158, opened 2018-10-12
swift-html

stephencelis:
Preliminary pull request to deprecate the Html module from this lib.


Tests/HtmlPrettyPrintTests/__Snapshots__/HtmlPrettyPrintTests/testPretty.1.html
```diff
@@ -13,23 +11,17 @@
                   class2
                   class3"
            style="color: red;
-                  background: blue;
-                  padding: rem(2);">
+                 background: blue;
```
mbrandonw: just seeing this. looks like we should add one more more space to the indentation...

PR #163, opened 2018-10-28
Allow customize default headers from NIO runner.

mbrandonw:
We had some annoying moments trying to customize CORS headers the past few days, so I'm surfacing the default headers here so that we can make those changes from inside pointfreeco.


Sources/HttpPipeline/NIO.swift
```diff
@@ -6,6 +6,14 @@ import Prelude
 
 public func run(
   _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data>,
+  defaultHeaders: [Response.Header] = [
+  .init("Referrer-Policy", "strict-origin-when-cross-origin"),
+  .init("X-Content-Type-Options", "nosniff"),
+  .init("X-Download-Options", "noopen"),
+  .init("X-Frame-Options", "SAMEORIGIN"),
+  .init("X-Permitted-Cross-Domain-Policies", "none"),
+  .init("X-XSS-Protection", "1; mode=block"),
```
stephencelis: Only concern is if folks set default headers they immediately lose a bunch of security. Not sure how to address short of a code-gen'd default like Vapor and Rails do, though.

stephencelis:
Closing for now. May revisit later.

PR #165, opened 2018-11-04
Update swift-snapshot-testing.

mbrandonw:


Sources/CssTestSupport/CssTestSupport.swift
```diff
@@ -1,9 +1,9 @@
 import Css
 import SnapshotTesting
 
-extension Strategy {
+extension Strategy where A == Stylesheet, B == String {
   public static func css(_ config: Config) -> Strategy<Stylesheet, String> {
-    var css = Strategy.lines.pullback { (stylesheet: Stylesheet) in
+    var css = SimpleStrategy.lines.pullback { (stylesheet: Stylesheet) in
```
stephencelis: Oh huh. This compiles without qualification?

mbrandonw: yeah the simple strategies do, guess it can figure it out if there's only one generic to deal with?

PR #174, opened 2018-12-19
Bool form decoder fixes

stephencelis:
This should allow for `1` to decode as true as well.


Sources/UrlFormEncoding/UrlFormDecoder.swift
```diff
@@ -324,7 +324,7 @@ public final class UrlFormDecoder: Decoder {
     }
 
     mutating func decode(_ type: Bool.Type) throws -> Bool {
```
stephencelis: I think we can avoid some of this repetition, but my first stab didn't work. Think the nested container logic may not be completely right.

PR #177, opened 2019-01-14
Swift 5

stephencelis:
A branch to track Swift 5 changes to the web.


Sources/HttpPipeline/NIO.swift
```diff
@@ -1,28 +1,29 @@
...
 import NIO
 import NIOHTTP1
+import NIOHTTPCompression
 import Optics
 import Prelude
 
 public func run(
   _ middleware: @escaping Middleware<StatusLineOpen, ResponseEnded, Prelude.Unit, Data>,
   on port: Int = 8080,
+  eventLoopGroup: EventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount),
```
mbrandonw: do we need this anymore? this was for the nio Postgres stuff right?

stephencelis: I think it's good to leave in for customization.

stephencelis:
Had to disable iOS tests for now :/

We can revisit, but given the experimental nature of this library I just wanted to land Swift 5.

PR #183, opened 2019-02-25
Close Channel on Error

### LINKED ISSUE #182: swift-web should close the Channel if an error is received in the last ChannelHandler
In SwiftNIO it's generally a good idea that on `errorCaught`, the `Channel` should be closed to prevent leaking network connections that are in some error states. In swift-web this should be added to  `private final class Handler: ChannelInboundHandler`. But this advice applies to any NIO ChannelPipeline: Generally the last handler (I usually call it the 'application handler') should close on unknown error.

Why? Generally, `ChannelHandler`s fire errors in cases where they don't know how to recover themselves. If another, later `ChannelHandler` knows how to recover, that `ChannelHandler` would consume the error. In other words: If an error reaches the end of the pipeline, nobody knew how to handle that error and closing the `Channel` is a good default.

This has always been true in SwiftNIO but in NIO 2 this will become especially important because `ByteToMessageDecoder`s (and the HTTP decoders) won't close the connections themselves anymore. Why is that? If the decoders themselves close the connection, then there's no way for the user to opt out.

The code to implement this is very straightforward:

```swift
public func errorCaught(ctx: ChannelHandlerContext, error: Error) {
    switch error {
    case let e as SomeErrorTypeIKnowHowToHandle where e.someCondition == expectedValue:
        // handle error
    case let e as SomeOtherErrorIKnowHowToHandle:
        // handle that error
    default:
        // log error?
        ctx.close(promise: nil)
    }
}
```

stephencelis:
Fixes #182.

We probably want to add some kind of logging, too.

### DIFF
--- a/Sources/HttpPipeline/NIO.swift
+++ b/Sources/HttpPipeline/NIO.swift
@@ -107,6 +107,10 @@ private final class Handler: ChannelInboundHandler {
       }
     }
   }
+
+  func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+    _ = ctx.close()
+  }
 }
 
 private func method(from method: HTTPMethod) -> String {
--- a/Sources/HttpPipeline/NIO.swift
+++ b/Sources/HttpPipeline/NIO.swift
@@ -74,11 +74,16 @@ private final class Handler: ChannelInboundHandler {
       }
     case .end:
       guard let req = self.request else {
-        _ = ctx.channel.write(HTTPServerResponsePart.head(HTTPResponseHead(
-          version: .init(major: 1, minor: 1),
-          status: .init(statusCode: 307),
-          headers: .init([("location", self.baseUrl.absoluteString)])
-        )))
+        ctx.channel.write(
+          HTTPServerResponsePart.head(
+            HTTPResponseHead(
+              version: .init(major: 1, minor: 1),
+              status: .init(statusCode: 307),
+              headers: .init([("location", self.baseUrl.absoluteString)])
+            )
+          ),
+          promise: nil
+        )
         _ = ctx.channel.writeAndFlush(HTTPServerResponsePart.end(nil)).then {
           ctx.channel.close()
         }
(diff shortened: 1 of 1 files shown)


Sources/HttpPipeline/NIO.swift
```diff
@@ -107,6 +107,10 @@ private final class Handler: ChannelInboundHandler {
       }
     }
   }
+
+  func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+    _ = ctx.close()
```
weissi: if you use `ctx.close(promise: nil)` you can save one allocation for when you don't care about the result anyway (I realise my example code didn't do that ;) )

stephencelis: Good to know, thanks! We've only started to dig deeper into NIO from our current lightweight wrapper. We've got a branch going that redefines middleware as `(Conn<I, A>) -> EventLoopFuture<Conn<J, B>>` everywhere instead of our current stack built on `IO`: https://github.com/pointfreeco/swift-web/compare/nio-conn

Haven't been able to fully flesh it out yet, though!

weissi: wow, nice. That sounds awesome!

Sources/HttpPipeline/NIO.swift
```diff
@@ -74,11 +74,16 @@ private final class Handler: ChannelInboundHandler {
...
-        )))
+        ctx.channel.write(
+          HTTPServerResponsePart.head(
+            HTTPResponseHead(
+              version: .init(major: 1, minor: 1),
+              status: .init(statusCode: 307),
+              headers: .init([("location", self.baseUrl.absoluteString)])
+            )
+          ),
+          promise: nil
```
stephencelis: @weissi Is it safe to assume that we should favor these `promise: nil` overloads over `_ =` wherever we're not `then`-chaining?

weissi: yes, there's no difference apart from allocating the promise/future that you'd then throw away with `_ = `. Therefore if you don't need the result, just passing `promise: nil` is just a bit cheaper. Here's the NIO code:

https://github.com/apple/swift-nio/blob/6050c9c864d641f66114cfd6ab690f1aa51d74d4/Sources/NIO/ChannelInvoker.swift#L141-L144

PR #186, opened 2019-04-01
allow basic auth even if header is lowercased

emilwojtaszek:
sometimes authorization header on Linux is typed lowercased which cause basic authorization fails, this simple fix patch the problem

### DIFF
--- a/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
+++ b/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
@@ -160,7 +160,7 @@ public func requireHttps<A>(allowedInsecureHosts: [String])
 
 public func validateBasicAuth(user: String, password: String, request: URLRequest) -> Bool {
 
-  let auth = request.allHTTPHeaderFields?.first(where: { $0.key == "Authorization" })?.value ?? ""
+  let auth = request.allHTTPHeaderFields?.first(where: { $0.key.lowercased() == "authorization" })?.value ?? ""
 
   let parts = Foundation.Data(base64Encoded: String(auth.dropFirst(6)))
     .map { String(decoding: $0, as: UTF8.self) }
--- a/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
+++ b/Sources/HttpPipeline/SharedMiddlewareTransformers.swift
@@ -160,7 +160,7 @@ public func requireHttps<A>(allowedInsecureHosts: [String])
 
 public func validateBasicAuth(user: String, password: String, request: URLRequest) -> Bool {
 
-  let auth = request.allHTTPHeaderFields?.first(where: { $0.key.lowercased() == "authorization" })?.value ?? ""
+  let auth = request.value(forHTTPHeaderField: "Authorization") ?? ""
 
   let parts = Foundation.Data(base64Encoded: String(auth.dropFirst(6)))
     .map { String(decoding: $0, as: UTF8.self) }
--- a/Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift
+++ b/Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift
@@ -183,7 +183,7 @@ class SharedMiddlewareTransformersTests: SnapshotTestCase {
     assertSnapshot(matching: log, as: .dump)
   }
   
-  func testBasiAuthValidationIsCaseInsensitive() {
+  func testBasicAuthValidationIsCaseInsensitive() {
     let urlRequestWithUppercaseAuthorizationHeader = URLRequest(url: URL(string: "/")!)
       |> \.allHTTPHeaderFields .~ ["Authorization": "Basic SGVsbG86V29ybGQ="]
     XCTAssertTrue(validateBasicAuth(user: "Hello", password: "World", request: urlRequestWithUppercaseAuthorizationHeader))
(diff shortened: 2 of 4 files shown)


emilwojtaszek:
@stephencelis will those tests fire automatically?

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
```diff
@@ -160,7 +160,7 @@ public func requireHttps<A>(allowedInsecureHosts: [String])
 
 public func validateBasicAuth(user: String, password: String, request: URLRequest) -> Bool {
 
-  let auth = request.allHTTPHeaderFields?.first(where: { $0.key == "Authorization" })?.value ?? ""
+  let auth = request.allHTTPHeaderFields?.first(where: { $0.key.lowercased() == "authorization" })?.value ?? ""
```
stephencelis: One change we could make is to use the `URLRequest` API for fetching this value, which appears to not worry about case.

``` swift
req.addValue("hello", forHTTPHeaderField: "Authorization")
req.value(forHTTPHeaderField: "authorization") // "hello"
```

```suggestion
  let auth = request.value(forHTTPHeaderField: "Authorization") ?? ""
```

Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift
```diff
@@ -68,6 +68,21 @@ class SharedMiddlewareTransformersTests: SnapshotTestCase {
     assertSnapshot(matching: middleware(conn).perform(), as: .conn)
   }
 
+  func testBasicAuth_Authorized_LowercasedHeaderName() {
```
stephencelis: > @stephencelis will those tests fire automatically?

@emilwojtaszek They will on macOS! But could you run `make linux-main` from the command line root of the repo to make sure they run on Linux?

emilwojtaszek: sure

Sources/HttpPipeline/SharedMiddlewareTransformers.swift
emilwojtaszek: looks much nicer!

emilwojtaszek:
looks like the test is failing, let me try if that patch fixes my problem first

mbrandonw:
it's a snapshot failure, so re-recording should set you straight!

```/home/travis/build/pointfreeco/swift-web/Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift:83: error: SharedMiddlewareTransformersTests.testBasicAuth_Authorized_LowercasedHeaderName : failed - Snapshot does not match reference.
@−
"/home/travis/build/pointfreeco/swift-web/Tests/HttpPipelineTests/__Snapshots__/SharedMiddlewareTransformersTests/testBasicAuth_Authorized_LowercasedHeaderName.1.Conn.txt"
@+
"/tmp/SharedMiddlewareTransformersTests/testBasicAuth_Authorized_LowercasedHeaderName.1.Conn.txt"
@@ −1,9 +1,9 @@
 GET /
−Authorization: Basic SGVsbG86V29ybGQ=
+authorization: Basic SGVsbG86V29ybGQ=
 
 200 OK
 Content-Length: 19
 Content-Type: text/html; charset=utf-8
 
 <p>Hello, world</p>
```

emilwojtaszek:
@mbrandonw locally tests are passing

emilwojtaszek:
also, looks like this change fix my problem 🥳

mbrandonw:
> @mbrandonw locally tests are passing

are you testing on linux locally? the failure is linux only. you can run linux tests by doing `make test-linux`, but it takes a bit of time cause it uses docker.

I think we are seeing a platform specific problem here. on Mac it records with an uppercase 'A' and on linux it uses lowercase 'a'. We could split this one test into two that are run on each platform.

emilwojtaszek:
in that case not sure if snapshot tests are the best option and that's why I've added a simple unit test, please check and let me know

mbrandonw:
Yeah that way works for me too.

@stephencelis whatdya think?

stephencelis:
I'm good with it! It's nice to know that there's an inconsistency with Linux, but we oughta file a bug at bugs.swift.org.

Tests/HttpPipelineTests/SharedMiddlewareTransformersTests.swift
```diff
@@ -182,4 +182,14 @@ class SharedMiddlewareTransformersTests: SnapshotTestCase {
 
     assertSnapshot(matching: log, as: .dump)
   }
+  
+  func testBasiAuthValidationIsCaseInsensitive() {
```
stephencelis: Small typo:
```suggestion
  func testBasicAuthValidationIsCaseInsensitive() {
```

Tests/LinuxMain.swift
```diff
@@ -126,7 +126,8 @@ extension SharedMiddlewareTransformersTests {
     ("testRedirectUnrelatedHosts", testRedirectUnrelatedHosts),
     ("testRequireHerokuHttps", testRequireHerokuHttps),
     ("testRequireHttps", testRequireHttps),
-    ("testRequestLogger", testRequestLogger)
+    ("testRequestLogger", testRequestLogger),
+    ("testBasiAuthValidationIsCaseInsensitive", testBasiAuthValidationIsCaseInsensitive)
```
stephencelis: Will need to regenerate or update this, too:
```suggestion
    ("testBasicAuthValidationIsCaseInsensitive", testBasicAuthValidationIsCaseInsensitive)
```

PR #188, opened 2019-06-05
Update Prelude

stephencelis:
This updates Prelude to the one with fewer dependencies.


Sources/ApplicativeRouter/SomePartialIsos.swift
```diff
@@ -180,25 +180,6 @@ extension PartialIso where B: RawRepresentable, B.RawValue == A {
   }
 }
 
-public protocol TaggedType {
```
stephencelis: Moving this to Point-Free proper.

PR #193, opened 2020-01-08
Swift 5.1

stephencelis:
Finally.


Makefile
```diff
@@ -1,39 +1,13 @@
...
 xcodeproj:
-	xcrun --toolchain swift swift package generate-xcodeproj --xcconfig-overrides Development.xcconfig
-	xed .
-
-bootstrap: common-crypto-mm xcodeproj xcodeproj-mm
-
-common-crypto-mm:
-	-@sudo mkdir -p "$(COMMON_CRYPTO_PATH)"
-	-@echo "$$COMMON_CRYPTO_MODULE_MAP" | sudo tee "$(COMMON_CRYPTO_MODULE_MAP_PATH)" > /dev/null
+	xcrun --toolchain swift swift package generate-xcodeproj
```
mbrandonw: think we could kill this too?

stephencelis: Same as other...needed for running iOS and macOS tests from the command line...

PR #198, opened 2020-09-10
Parse headers for requests.

mbrandonw:
I think we'll need to parse stuff from headers soon and apparently this parser doesn't support it yet...


stephencelis (approved):
nice. surprised we didn't have this!

PR #202, opened 2021-10-04
Prepare to remove BlueCryptor

stephencelis:
This does some preparatory work to migrate from BlueCryptor to swift-crypto. swift-crypto does not include the older version of AES encryption we were doing, so this moves to the new AES encryption that is supported for fresh encrypts. And for decrypts it will fall back to BlueCryptor.

This way we can roll the change out and let active folks slowly get updated cookies so we don't log them out. Once we fully remove BlueCryptor, those with stale cookies will be logged out, but it's not a huge deal since logging in is a quick GitHub redirect.


Package.swift
```diff
@@ -4,7 +4,7 @@ import PackageDescription
 let package = Package(
   name: "swift-web",
   platforms: [
-    .macOS(.v10_13),
+    .macOS(.v10_15),
```
stephencelis, line 7: Minimum deployment target for CryptoKit.

mbrandonw (approved):

